PR #310, opened 2015-05-09
New fragment-based Preferences system

baka-kaba:
This is an overhaul of the whole system, using fragments instead
of PreferenceScreens. It has a dynamic single/dual-pane system that
works similarly to preference headers, except you can use normal
preference layouts instead of only submenu headers.

Each subscreen is defined by its own fragment - the system includes
some basic behaviour baked in (automatic summary handling, launching
new fragments on a click etc.) and is easily extensible if you want
to add custom functionality for things like updates and listeners.

It uses the app's theme and updates immediately when the user
picks a new one, and also the SwitchPreferences are finally here!


baka-kaba:
I finally got to it since you bumped up to API 14 - it turned out that moving to fragments meant writing a whooooole lot of framework if you didn't just want basic headers, and I had to ditch PreferenceScreens to get everything opening in the same activity and looking nice. Knowing my luck there's probably a library that does all this but hey, I gave it a go! Hopefully it all makes sense

Seems to all be working anyway, I've tested on 22 on an N5, and 14 on an emulated GNex and 16 on an emulated N7.

Sereri:
I'm guessing there is no alternative to having 9 different fragments?

baka-kaba:
Not that I can think of - each one represents a different settings screen, and to have a dual-pane mode you need to be able to swap out the content of the right pane. And each fragment needs a separate XML file, unless there's some trickery I haven't seen yet. I could probably cram a few fragment classes together into a single file, but I wanted to avoid some complex monolithic file that's hard to work with or add to, y'know? I think this is pretty straightforward once you get the idea.

It might be possible to have simple fragments (that just have a layout) created programmatically, and use reflection or a lookup to see if there's an actual fragment class to instantiate instead. That way you'd only need to create a fragment class if you needed to add some custom behaviour. I'm already looking at automatically adding linking behaviour by checking if there's an `android:fragment` key in a preference's xml (instead of adding it to the lookup array by hand) so I might be able to refactor it a bit, I just want to make sure it's as light as possible so people can add things nice and easily. And make sure it's working for everyone too

What's making you uneasy about it? Is it the fact it's pretty different to the usual way of managing them, or the number of classes it adds, or something like that? Don't hold back here, I tried to build something that acts the same way and is simple to work with, but feedback helps! I thought moving to fragments was going to be fairly simple but... nope. Preference-headers would work, except the root page can *only* contain links to subsections (like in the Youtube app), so we'd have lost all the category formatting and elements you wanted to keep on the main page, like the About button and Import/Export settings

Sereri:
Eh, I was just hoping it would be possible to split it into less parts, more or less the way it was. I guess that this is probably the best way though. I'll give this a spin and merge it in if nothing breaks spectacularly.

baka-kaba:
Yeah I hoped that too... and ended up having to write a whole system to handle it. It was a pain but I think it handles pretty nicely now, and I definitely think I can simplify things a bit here and there - I'd like to have more stuff just defined in the XML. And I mean, I'm learning as I go with a lot of this, so there's probably a bunch of things I should have done differently, but we'll see!

Sereri:
I can't actually just merge it, did you pull the latest changes from AniMD?

baka-kaba:
Oh ok, it just works huh - there we go

PR #318, opened 2015-05-15
Add number picker dialog for Posts per Page preference

baka-kaba:
I borrowed the dialog used in `displayPagePicker()` in `ForumDisplayFragment`, so we get a nice spinner deal instead of having to type in a number and wrangle it internally. Also feel free to set a different minimum - I'm guessing the forums don't actually allow more than 40 per page? I can't do it on the desktop anyway, and it causes issues in the app if you try

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/preferences/AwfulPreferences.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/preferences/AwfulPreferences.java
@@ -243,7 +243,7 @@ public class AwfulPreferences implements OnSharedPreferenceChangeListener {
         avatarsEnabled           = mPrefs.getBoolean("avatars_enabled", true);
         hideOldImages            = mPrefs.getBoolean("hide_read_images", false);
         showSmilies              = mPrefs.getBoolean("show_smilies", true);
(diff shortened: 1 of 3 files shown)


baka-kaba:
Also I don't know if an issue needs opening for this, but I noticed some issues on my N7 emulator (API 14) when you change the posts-per-page value - I saw them when I was testing, but I reverted to an earlier commit and they've been there the whole time, so it's a long-running bug I guess.

The big issue I've only seen on the N7 emulator - open a thread, go to settings and change the PPP value, then come back to the thread and hit the back button. It tries to go back to the forums/threads split view, but the fragment transition gets stuck, like this:
capture
Can anyone check this on a real device, or an emulator that's not running on a bad computer without virtualisation?

The other problem is if you have a thread open when you go into settings and change the PPP value, when you come back the prev/next page buttons don't do anything until you reopen the thread. I think it's when the 'current' page no longer exists (like when you go from a low PPP value with more pages to a high PPP with few pages) so the prev and next pages don't exist either. The page numbers don't update though, so it still looks like you're on page 140/170 or something valid, when the real last page is now 130 with your new settings. I've seen this on my Nexus 5 too.

PR #326, opened 2015-05-25
Rework thread rating handling and 'fix' the Film Dump (fixes #324)

### LINKED ISSUE #324: Fix the film dump
Film dump has no thread tags but ratings, everything looks terrible. Introduce rating tags from awful thread tags.

baka-kaba:
I've broken out the ratings handling code into a new utility class
that handles all the parsing and drawables, and wrangles the different
types of ratings on the site. Hopefully it'll be easier to add to if
this crops up again or if they change the URLs for the different rating
types.

It also lets you check the type of rating you're looking at, so things
can be handled in code if necessary - and it's doing that right now for
the amazing 'make the empty thread icon view visible' hack when you're
looking at film dump threads. It's a stopgap until... someone makes it
look nicer, but it works and hopefully it'll make it easier to manage
the layouts dynamically.

Also I added the sweet star icons from the site, feel free to replace
the hell out of them


Sereri:
Can you rework it by using 0.0 -> 5.0stars.png from the https://github.com/Awful/thread-tags and put set it as the thread tag in the view?

baka-kaba:
You got it

Just one thing to check - even though they look like special thread tags in the Film Dump, they're still technically ratings, so their visibilty is controlled by the Show Ratings preference and not Show Thread Tags. Personally that's what I'd expect, so seems good to me, but I just thought I'd check

baka-kaba:
You know what, I was looking at secondary tags and I realised I could handle them the exact same way, so I could pull out most of the code into a superclass and have AwfulRatings and AwfulExtraTags or whatever extend that (and anything else that comes along). So a general 'put in a URL and get a reference to the thing it represents' system

I'll give it a go and see what you think

baka-kaba:
Ok done, secondary tags are in and stuck on the thread tags as a little button. That probably needs a designer to take a look!

I probably overdid it a bit with the code, but I wanted to break it out of that huge class and make it extensible in case anything else is ever added, or they mess around with the URLs or whatever. It's doing the job anyway

PR #335, opened 2015-06-02
Switch nav drawer to support Design lib version

baka-kaba:
(I know you already fixed this, but I'd already started playing with it :frog: )

I've reworked the navigation stuff to use the new Design library, it's a bit easier
to work with too (mostly defined in XML with a callback for item selection).
I added a material-ish header without an actual graphic (just a horrible blown-up
app icon with a tint because why not), and a mini avatar icon. Pre-lollipop will
need some library or custom code to make it round.

Forum/thread names update immediately now (well, as soon as the loader is done)
I had to make a choice about exactly what happens with this stuff, since you
can have a thread open (with an associated forum) and then visit a different
forum in its fragment without opening a thread. So you end up with a forum
and a thread that are unconnected, so which do you show in the menu?

I went with the last fragment to update getting to set the details - it's all handled
in that setNavIds method, so it could be tweaked with a call to that whenever a
fragment is in view, or something. The loaders are kind of all over the place so it's
still inconsistent, things triggering in the background and re-updating 'what you're
looking at', but that's always been an issue!

### DIFF
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -43,22 +43,23 @@ android {
 }
 
 dependencies {
-        compile 'com.android.support:appcompat-v7:21.+'
-        compile project(':libraries:tree-view-list-android')
-        compile 'com.mcxiaoke.volley:library:1.+'
-        compile 'com.github.androidquery:androidquery:0.26.8'
-        compile 'com.google.code.gson:gson:2.3.1'
-        compile 'org.jsoup:jsoup:1.8.2'
-        compile 'joda-time:joda-time:2.7'
-        compile 'com.samskivert:jmustache:1.10'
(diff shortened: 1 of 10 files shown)


baka-kaba:
Also you might need to do Build > Clean Project if you get 'what's support-v4, what's a fragment' issues after adding the design library. Might be stating the obvious but it confused me!

It's possible to do highlighting on the nav menu options, to show what section you're in, but since you're sort of in multiple sections at once (forum list, thread/bookmarks list, a thread) it's tricky to know what to show. I'll have a go at making it reflect the state of the menu (like whether it's showing a forum name or a thread) but like I said in the commit, that's pretty inconsistent right now because of all the loaders jumping in saying HEY I JUST UPDATED THIS FRAGMENT

PR #394, opened 2016-03-23
Add changelogs

baka-kaba:
Added a new dialog which is automatically shown when the version mode changes
(unless it's the first run, where you get the GIF warning etc), and is also an option in
the settings menu.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.java
@@ -36,6 +36,7 @@ import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.drawable.ColorDrawable;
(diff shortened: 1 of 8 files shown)


Sereri:
Could maybe use an option to "never show again" or something like that, maybe as the negative button on the dialog, with a setting somewhere to reenable it (yay more settings)

baka-kaba:
You mean so people never see the changelogs again unless they opt back in? I could do, wouldn't be hard, but I feel like it might be better if everyone has to see it. It only happens when they update, which isn't _that_ often, and then they get informed about what's been added. That's the reason I added it really, a lot of people seem unaware of things like embedded tweets or other changes, and then you get loads of questions in the thread about stuff that's already been done but nobody's using.

I'll add a flag and a switch if you like, it just seems like this is something where making people :dealwithit: would bring the most benefits. What do you think about adding it as-is, and if enough people complain we can make it optional?

I think **What's new?** might be a better title for the dialog too, a bit friendlier and intriguing than CHANGELOG

Sereri:
Yeah Ok

As for the title, I think the settings button is alright, but What's new might be better as the dialog title, yeah

baka-kaba:
All right, I did the thing

PR #397, opened 2016-03-27
Rework prefs to use type-validated references instead of string literals

baka-kaba:
I've centralised all the preference keys as string resources, so everything
can refer to a single variable instead of using string literals. I didn't like
how brittle it was to have copies of the same string that are at risk of typos
and hard to refactor, and the way the rest of the app needed to know the internals
of how preferences are handled.

The preference keys are stored as resource strings (so the settings XML can
refer to them), and there's a set of constants in the Keys file which the app
can use. Those are passed to set and get methods in AwfulPreferences, and they're
grouped in typedefs according to the type of preference they set (int, boolean etc).
So you can set any preference by its key constant, and passing the correct type is
enforced by the typedef-annotated method.

Hopefully this'll avoid any future bugs where someone accidentally mistypes something
or uses the wrong type and then who knows what's going on in there

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulLoginActivity.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulLoginActivity.java
@@ -32,7 +32,6 @@ package com.ferg.awfulapp;
 import android.app.Activity;
 import android.app.ProgressDialog;
 import android.graphics.drawable.AnimationDrawable;
-import android.os.AsyncTask;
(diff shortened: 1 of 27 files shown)


baka-kaba:
I'm not 100% on whether the constants need to be there, or whether it's better to just pass the resource IDs? It's easy enough to inline them and the values will still be in the typedefs, so the validation will still be there. I added them with literals before I realised the keys would need to go in R anyway, because of the XML needing to reference them, so maybe they could go?

PR #398, opened 2016-03-28
Tweak the post actions layouts to fit Material guidelines

baka-kaba:
Basically I've
- shrunk the icons to 24dp
- adjusted the margins and padding to fit the
  Simple Dialog keyline recommendations
- changed the width to the 5 * 56dp minimum (280dp)
- added the minimum margins around the dialog
- tried to adjust the font sizes except they don't list those anywhere

### DIFF
--- a/Awful.apk/src/main/res/layout/action_item.xml
+++ b/Awful.apk/src/main/res/layout/action_item.xml
@@ -3,23 +3,26 @@
     android:layout_width="fill_parent"
     android:layout_height="wrap_content"
     android:minHeight="64dp"
-    android:orientation="horizontal">
+    android:orientation="horizontal"
+              android:paddingRight="24dp">
 
     <ImageView
         android:id="@+id/actionTag"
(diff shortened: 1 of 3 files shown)


Sereri:
Finally had the time to take a look at this
Not sure why the icons are so small. The account icons in the simple dialog keyline mockup, while not explicitly labeled as such, are 40dp big (specs, almost all the way down)

baka-kaba:
Yeah, but those are the avatar ones from here:
[](https://www.google.com/design/spec/layout/metrics-keylines.html#metrics-keylines-touch-target-size)
The icon one is 24dp. Honestly that's what made me look into the material guidelines in the first place, I opened the menu and the icons were all whoa huge

But I still don't think I've got the title/content spacing quite right - and honestly, the menu in Chrome is a lot more compact, the url title is a much smaller font and that could be adapted with icons too. It's up to you really, you added it! I think it looks better with the sizes reduced though, especially the icons

PR #403, opened 2016-04-02
Fix URL menu handling with missing path parts

baka-kaba:
The image/gif checking code was assuming a last path segment exists,
 and causing NPEs when they didn't (e.g. for http://www.google.com)
(cherry picked from commit c85e78c)


baka-kaba:
I just realised Sereri fixed this yesterday...

I cleaned up the code a little bit with some comments, if that helps! Otherwise I'll close it

Sereri:
I fixed it but you _fixed_ it

PR #404, opened 2016-04-03
Fix vector resources for < API 21

baka-kaba:


baka-kaba:
Uh that probation issue was kind of a problem (for bad dudes)
Affected people will need to clear data to fix that I think

Issue #349, opened 2015-06-23
Square forum tags?


baka-kaba:
Do we want this for the forum list? I tried it out and it's easy to implement - I used the `ElementCollection` thing, like with the review/secondary tags, it just needs some... square tags from somewhere. There are 3 or 4 already - do we have any anywhere else? It would match the rest of the app better and give a bit more room for the titles

For a laugh I tried adding a theme-dependent tinting call on the `getDrawable()` call, so they're all green for yospos, yellow for amberpos etc (it uses the colour from the theme). You can have a look at it [here](https://github.com/baka-kaba/Awful.apk/tree/ForumTags) if you like, it's bad inefficient code but I was just trying it out quickly

It only affects assets accessed through that `ElementCollection` class, so not normal tags, but I could add them to a Tags class if people are cool with it? It's a different way of accessing things (calling the static class with the URL instead of parsing a filename out and getting `R.drawable.filename`), but it should be easier to maintain, provide defaults if they ever change a URL, that kind of thing

Sereri:
Problem is, the tags were now using are a byproduct of the Awful iOS app. They don't use forum tags so there are only those where the forum tag is also a thread tag (like GiP).

I mean we could ask @diabolik900 and @Ridilla, but at the moment I don't think it's possible

diabolik900:
Making forums tags in the style of our square thread tags should theoretically be easier than the thread tags because they all follow the same simple style. 

I have some free time coming up over the weekend where I could probably work on them if there's interest.

baka-kaba:
It's pretty much ready to go if you're willing to make them, yeah! And if Sereri's cool with the change

If you wanted we could incorporate the frog silhouette somehow for the two-tone pattern, like taking up the bottom/right half of the square, to help give the app a little more character. I'm not a designer in case you hadn't guessed!

Also a generic icon would be useful too, as a placeholder for when new forums are added or something's changed. We can just have it use the icon from the webpage if necessary, but it'll look pretty rubbish

Sereri:
Some time has passed and I made a generic forum tag that works by tinting the white for the square color and setting the background of the imageview to the color of the background (good job on github having a white background)
![forumtagtest](https://cloud.githubusercontent.com/assets/803440/14994898/752bd62c-1172-11e6-8440-25005f374123.png)

Problem is it looks bad.
[...]

PR #418, opened 2016-05-16
Create Forums package with update/sync tasks, new forum index

baka-kaba:
Added a Forums package responsible for updating and providing access
to forum data.

ForumRepository handles async updates with callbacks, handles database actions,
and provides access to data as ForumStructures.
ForumStructure is a representation of the forum hierarchy, which can produce
customisable list and tree formats.
Forum encapsulates the data associated with a forum, with some utility functions.

Also reworked the forum index page - now it uses a RecyclerView and the
new ForumStructure system to make a customisable list. There's a TagProvider class
which handles setting forum tag views, and a SquareForumTag class which is a
specific implementation using Sereri's graphic and with a dynamic text overlay.

Added a new settings category for the forums list, and with a force update option.

Also included a SyncManager class to run necessary update/sync tasks on app start
and login. Forum data updates once a week at low priority, unless there's no data,
in which case there's a high priority update and a basic fallback update if it fails.

Upgraded database version to force a forum data clear, since old data will
look messed up with the new Forums system data

Data limiting uses the user's 'download images on 3G' setting to restrict
low-priority forum updates. This could probably do with a proper setting
for limiting data in general, maybe as a single option or with granular settings.

Basic square thread tags, generated
(cherry picked from commit 84bea58)


Sereri:
Looking very good (shame about having to hardcode all the forum shortnames but whaddayagonnado)

I think there might be a small bug when the parsing is finished but the view is not visible. The view just keeps on "loading" until the app is fully closed and reopened.

Sereri:
By the way, I'm almost instantly turned off section headers and subtitles (because I hate change). Looking very nice either way.

baka-kaba:
That's why I made them optional! To avoid complai- I mean, because freedom of choice is good! (It's still gonna happen though, remember when thread items got slightly taller...)

Good catch on the refresh bug, thanks - I had it refreshing in onResume before, but that reset the index every time and I moved it earlier in the lifecycle. Looks like I forgot what else happens! Looks like it's only when the activity isn't visible (like if you're in Settings), I've fixed it up and tweaked a few other things too

Sereri:
Ok, I think this is almost ready. Looking through the changes I think we should rename forumtagtest.png to something else.

Speaking of which, I also tried another one without the black borders and I like it a little bit better:

forumtagtest_new

What do you think? I still think it could use a bit of texture like the thread tags (see computers, misogyny or comedy) but I'm terrible at photoshop etc. Maybe we can get @diabolik900 or someone else from the Awful.app guys to hook us up.

diabolik900:
Happy to lend a hand if I can, but I can't see your image. I assume it's the same forumtagtest.png image that you mentioned earlier in your comment?

Sereri:
That would be great.
And I removed the black lines so it's only white and transparent now, with the white github background there's not much to see here. Try clicking on the empty space.

Anyway, I'm gonna merge this and shift the discussion back to #349

PR #429, opened 2016-05-27
Update launcher icon, fix https PM intent

baka-kaba:
- remade icons up to xxxhdpi, moved them into mipmap
  (see: http://android-developers.blogspot.co.uk/2014/10/getting-your-apps-ready-for-nexus-6-and.html )
- fixed the private messages intent only handling the http scheme


Sereri:
I think I broke this by merging in the wrong order perhaps

Sereri:
Oh I get it, both pull-requests seem to have the "fixed the private messages intent only handling the http scheme" part

baka-kaba:
Whoops, that's what I get for breaking a commit up by hand

PR #432, opened 2016-05-29
Clean up thread list item layout

baka-kaba:
Removed the v19 version and reworked the generic one so it works everywhere,
mostly pulling out the fixes from #7c0d48ec and applying them as version-dependent
dimension values. Also stuck moved some constant attributes out of code and into
the layout file, and reused the text styles from the forum index so they match

Added some tooling attributes so we can actually see the layout in the editor (I just found
out about this, it's amazing), and cleaned out some old values


baka-kaba:
I changed the unread count to use sp instead of dp because the linter was complaining, I don't know if you want to tweak that since it looks pretty different on my screen - not in a bad way, I just use small text and really it's text that should scale for people right?

Are we sticking with the right-side screen padding at 4dp? I know it's a design choice to make a little more space but it still feels a bit weird

Sereri:
Wow I had no idea you could do that

I guess I should have made the changes the default layout to begin with

I don't really see a difference with the sp/dp change so v:neutral_face:v

I'd rather like to keep the right 4dp, it looks a bit squished if you remove the padding imho

baka-kaba:
Oh I meant _more_ padding... like really it's meant to be 16dp on both sides, but it's pretty dense as is with the title so I get if you want to keep it

If you have your font size set to normal then dp=sp as far as I'm aware, it's a little smaller on mine now but that's totally cool

Your layout voodoo with that negative margin stuff is pretty whoa, by the way. I was fighting with that layout and trying to get everything working the same way without weird alignment bugs, and that seems like pretty much the only way to do it... of course I didn't look in the git history first or anything

Issue #433, opened 2016-06-07
Title regression


Sereri:
The current beta build has a regression with the title. When returning to the thread view for example the correct title will appear, then be replaced by the title of the forum display view

![screenshot_20160607-104236](https://cloud.githubusercontent.com/assets/803440/15851368/b9423640-2c9c-11e6-9434-0938ccd2a736.png)

baka-kaba:
Fixed by #583 (fragments can't push titles anymore, they notify and the view pager decides on a title based on the current page)

PR #434, opened 2016-06-23
Mitigated race condition on phones

Sereri:
Possible fix for #433 (on phones)

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulFragment.java
@@ -207,7 +207,7 @@ public abstract class AwfulFragment extends Fragment implements AwfulRequest.Pro
         // TODO: fix race condition in ForumDisplayFragment and ThreadDisplayFragment - both restart their loaders in onResume,
         // both of those set the actionbar title - even in phone mode where only one is visible. Whichever loads last sets the actionbar text
         AwfulActivity activity = getAwfulActivity();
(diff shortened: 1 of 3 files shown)


Sereri:
@baka-kaba I checked and the race condition still happens on tablets (where it could be checked which fragment is the first I guess) if both fragments are visible, but whatever comes first doesn't matter that much because is still technically true. Seems good?

baka-kaba:
Yeah, my idea for this was for the fragments to give the activity a title, instead of setting it directly, and then the activity can decide what title(s) to show based on what's visible. Feels like it makes more sense for the activity to be in control

Actually let me try something

baka-kaba:
Ok yeah, I've seen how it works in the activity. I think I'd just change the visibility check, move it out of `AwfulFragment` and put it in the `setActionbarTitle` method in `ForumsIndexFragment`. If you change line 769 from
`if (requestor instanceof AwfulFragment && ((AwfulFragment) requestor).isVisible()) {`
to
`if (requestor instanceof AwfulFragment && isFragmentVisible((AwfulFragment) requestor)) {`
that way the Activity's in control of what's being shown

I was trying to add some code that keeps track of the current title for each fragment, and having a title/subtitle thing going on in the action bar (like current forum/current thread), but the way it's done with the scrollable title it's all a bit awkward. I think that needs to be decided as part of a tablet mode overhaul really, so yeah just letting it change randomly is probably fine for now until there's a concrete idea about what should be shown

It should probably be tied into the navigation drawer updates too (it's updating with the same info, current titles) but yeah, needs some designing first

PR #448, opened 2016-10-23
Reply tweaks

baka-kaba:
Two things really - I've added some more advanced tagging functionality so it's cleaner and easier to work with (and now does all the tags and fixes youtube links from the Android client), and I've moved the tagging/emote options into a bottom sheet and changed the overall layout a bit


Sereri:
Looking very good

PR #451, opened 2016-11-07
Theme bottom sheet in reply view

baka-kaba:
Added some attributes for background and text colours.
Seems like they have to be applied programmatically, the library applies a default value
if you don't supply one, overwriting anything defined in the theme you can pass


Sereri:
Heh, I did the same thing earlier but ended up not commiting it yet waiting for https://github.com/rubensousa/BottomSheetBuilder/issues/9

PR #462, opened 2017-01-15
Rework theming and colour/CSS handling methods

baka-kaba:
Created new ThemeProvider enum class to define themes, and provide utility methods for accessing
related resources, defaults etc.
Reworked ColorProvider as an enum class with simplified methods, mapping constants to themed and
default colour attributes instead of a method for each with a switch for each theme.

Themes have been reworked to try and set some basic attributes and define other attributes'
relationships to them, so you can set some standard colours and certain elements will use them
unless the theme overrides it. It's probably still not very consistent and maybe it needs reworking,
but I think it's easier to understand now, once you get that some attributes are defined in a parent.
I cut out a bunch of redundant attributes too

I've also moved the SRL progress spinner colours into arrays, referenced directly by themes,
so they can be customised purely in XML. Bookmark colours too
Also tweaked some themes for more contrast, including some half-bright colours for YOSPOS. I dunno
know if that's canon or an outrage or what

This also fixes some issues where user css files weren't being checked pre-Marshmallow

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.java
@@ -6,7 +6,6 @@ import android.content.Intent;
 import android.graphics.Typeface;
 import android.net.http.HttpResponseCache;
 import android.os.Bundle;
-import android.os.Message;
 import android.support.v7.app.ActionBar;
 import android.support.v7.app.AppCompatActivity;
 import android.text.method.ScrollingMovementMethod;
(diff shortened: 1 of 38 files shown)


baka-kaba:
Wow reviews! Just thought I'd check this isn't too much messing around

Sereri:
Finally got around to check this out (oh god, I think I made this pun before). The whole Theming rework is pretty great, good job (as usual, really). As for the YOSPOS changes, I can't tell the difference really. Do you have an example of something that changed?

I also just experimented with dropping the restarting of the ForumIndexActivity ( lines 590-595). There used to be a bug in the support library when changing themes and using the ViewPager that caused crashes but it doesn't seem to happen anymore. Lots of coloring does get fucked up though.

Sereri:
You know what, nevermind, I'm probably just an idiot. I forgot to comment out the overridden method so it never gets recreated. I'll check tonight when I get back home but I'm sure it'll crash.

Sereri:
Turns out that while the coloring changes were indeed my fault (by not removing the overridden afterThemeChange as well), switching themes on the fly will no longer crash the app 

Well, I only tested as low as 4.4, all my AVDs with lower versions die for no apparent reason.

baka-kaba:
Cheers!

I added another commit - are those all the changes you meant? Let me know and I'll squash it and pull

Sereri:
Yeah, I think that's all.

PR #464, opened 2017-01-30
Extract WebView code to new AwfulWebView class and utilities

baka-kaba:
Pulled out most WebView configuration code into several components:
- an AwfulWebView class with some common methods and lifecycle handling
- a WebViewConfig class that does most of the setup that was previously
  in the containing fragments
- LoggingWebChromeClient, which is the basic client with logging calls,
  previously used in one class but now easy to subclass and drop in anywhere
- WebViewJsInterface, holding common JavascriptInterface-annotated methods

The idea is to get all that stuff out of the fragments holding the webviews,
get the code all in one place, and make the configuration and interaction calls
as minimal as possible.

There are still some changes to make (lots of TODOs),
partly removing some old config code and maybe pushing it all into AwfulWebView
as a generic init call, and partly getting some of the remaining WebView-poking
code out of the fragments (they shouldn't need to make direct load calls with
base URLs and html and charsets and stuff - need simple utility methods for that)

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/MessageFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/MessageFragment.java
@@ -1,6 +1,5 @@
 package com.ferg.awfulapp;
 
-import android.annotation.SuppressLint;
 import android.app.ProgressDialog;
 import android.content.ContentResolver;
 import android.content.ContentUris;
@@ -22,8 +21,6 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.view.ViewGroup;
-import android.webkit.WebSettings;
-import android.webkit.WebView;
 import android.widget.EditText;
 import android.widget.ImageButton;
 import android.widget.TextView;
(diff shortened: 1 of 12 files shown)


baka-kaba:
Am I all right merging this in?

I've been testing it a while, I'm pretty sure everything works as it did - might want to check the post previews though

Sereri:
I think so, seems alright. Will test it tonight.

I'm pretty sure we already had a AwfulWebview once but I think it was just a wrapper for ... some reason I don't remember and became redundant in some later version of android.

baka-kaba:
Yeah a file with that name was already being tracked by Git, spooky

There's a bunch of redundant code in there - I refactored all the setup code as methods in one class, and tried making all the webviews use the main setup method, which worked fine. So those old methods are basically legacy code and the class isn't really needed, we can just have all the setup stuff in AwfulWebView

I was thinking it might be easier to add another commit removing all that, so that if there *is* a problem it's easy to go back and find how things used to be set up for each component

baka-kaba:
Sorry, fixed that regression (that I artisanally merged by hand thanks git)

Sereri:
Seems to work fine in all 3 cases. Are the TODOs anything serious?

baka-kaba:
Nah they're either stuff that was in the original code (like WebView config warnings) or things to keep in mind for the future

PR #478, opened 2017-03-21
Added DownloadListener to webview in ThreadDisplayFragment

Sereri:
I hope this doesn't pollute your great webview refactoring 

I only added functionality for the TDF because ~it was so very easy that way~ downloading only makes sense to me in that context.


baka-kaba:
Looks good to me! I mean yeah, makes sense to add it there because it's specific to that fragment. I guess you could potentially have it in the PM view later - but that would mean refactoring the actual download code too, so worry about that later eh

For some reason on mobile it wants to open the 'Commits' tab in a video player 🤔

PR #480, opened 2017-04-14
Reply rework

baka-kaba:
Some tweaks and fixes for the Reply view:

- save/discard is a forced choice when the user cancels a reply
- automatically saves in onPause (avoids losing replies when the app is in the background etc)
- preview is now an option on the Confirm Submit dialog
- reorganised methods to try and make it easier to read, added documentation
- simplified/encapsulated some state
- some code fixes to try and avoid problems (e.g. activity-dependent code into onActivityCreated)


baka-kaba:
Hey Sereri, I've left this as a PR to make sure it doesn't interfere with anything you're doing

PR #484, opened 2017-04-27
Settings

baka-kaba:
I've moved the Embedding options to a submenu, tweaked some other wording, and fixed a DP/SP issue with the font size preview


baka-kaba:
This is ready to go, but I thought I'd run the wording changes by you to see if there's anything you want to change. It's basically
* title for the embedded stuff category
* description for that and the highlighting category, at the top
* a better name than 'General video' for all the WebMs and whatever
* the autoplay only applies to 'General video' right now (actually only .webm looking at the javascript) so does that need a wording change, or just keep it as 'video' in case we want to add other stuff?
and whatever else, I removed a bit of text to try and keep it simpler and avoid repeating things

Sereri:
Maybe "Convert video links" or "Embed video links" instead of "General video"?

baka-kaba:
Well it's in the Embedding section and under a Video subheading anyway - I'm really looking for a name to contrast with YouTube and Vimeo and Vine (rip). I want to avoid the word 'convert' too, I'm looking at putting in an option to actually convert gif links to video where possible

How about Other? The summary underneath says WebM and mp4 etc, that might be a good catch-all

Sereri:
Yeah I like "Other" the best 👍

Issue #458, opened 2016-11-26
Show filesize of images when clicking on them 


Sereri:
So that you can decide whether or not you want to load that 150mb gif

Sereri:
Did anything ever came of this?

baka-kaba:
This was rolled in, there's a related issue in #520 to expand handling for this

PR #487, opened 2017-05-02
Show Size of Images in Action Menu

Ephemerial:
Image size is now displayed at the top of the post action menu, within the title and before the image URL. 

In the future some form of indicator should be displayed to the user while image size is retrieved rather than apparent unresponsiveness while the network request is processed.

This PR is intended to resolve issue #458.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
@@ -202,9 +202,9 @@ public class PostActionsFragment extends AwfulDialogFragment {
 	public void changeTitle(String title) {
 		View view = getView();
 		if (view != null) {
-			TextView actionTitle = (TextView) view.findViewById(R.id.actionTitle);
-			if (actionTitle != null) {
-				actionTitle.setText(title);
+			TextView fileSize = (TextView) view.findViewById(R.id.fileSize);
+			if (fileSize != null) {
+				fileSize.setText(title);
 			}
 		}
 	}
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
@@ -198,6 +198,16 @@ public class PostActionsFragment extends AwfulDialogFragment {
 	public void setTitle(String title) {
 		this.title = title;
 	}
+	
+	public void changeTitle(String title) {
+		View view = getView();
+		if (view != null) {
+			TextView actionTitle = (TextView) view.findViewById(R.id.actionTitle);
+			if (actionTitle != null) {
+				actionTitle.setText(title);
+			}
+		}
+	}
 
 	public void setPostId(String postId) {
 		this.postId = postId;
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
@@ -205,6 +205,7 @@ public class PostActionsFragment extends AwfulDialogFragment {
 			TextView fileSize = (TextView) view.findViewById(R.id.fileSize);
 			if (fileSize != null) {
 				fileSize.setText(size);
+				fileSize.setVisibility(size == null ? View.GONE : View.VISIBLE);
 			}
 		}
 	}
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
@@ -199,12 +199,12 @@ public class PostActionsFragment extends AwfulDialogFragment {
 		this.title = title;
 	}
 	
-	public void changeTitle(String title) {
+	public void setSize(String size) {
 		View view = getView();
 		if (view != null) {
 			TextView fileSize = (TextView) view.findViewById(R.id.fileSize);
 			if (fileSize != null) {
-				fileSize.setText(title);
+				fileSize.setText(size);
 			}
 		}
 	}
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/PostActionsFragment.java
@@ -205,7 +205,7 @@ public class PostActionsFragment extends AwfulDialogFragment {
 			TextView fileSize = (TextView) view.findViewById(R.id.fileSize);
 			if (fileSize != null) {
 				fileSize.setText(size);
-				fileSize.setVisibility(size == null ? View.GONE : View.VISIBLE);
+				fileSize.setVisibility(size == null ? View.INVISIBLE : View.VISIBLE);
 			}
 		}
 	}
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
@@ -1238,8 +1238,7 @@ public class ThreadDisplayFragment extends AwfulFragment implements SwipyRefresh
 						URL location = new URL(parameters[0]);
 						URLConnection connection = location.openConnection();
 						int size = connection.getContentLength();
-						String pretty = Formatter.formatShortFileSize(getContext(), size);
-						return String.format("Image Size:\n%s\n\n%s", pretty, url);
+						return String.format("%s", Formatter.formatShortFileSize(getContext(), size));
 					}
 					catch (IOException exception) {
 						exception.printStackTrace();
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
@@ -60,6 +60,7 @@ import android.support.v4.content.Loader;
 import android.support.v4.view.MenuItemCompat;
 import android.support.v7.widget.ShareActionProvider;
 import android.text.TextUtils;
+import android.text.format.Formatter;
 import android.util.Log;
 import android.view.InflateException;
 import android.view.LayoutInflater;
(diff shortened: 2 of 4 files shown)


Sereri:
Hi Ephemerial,

thanks for your pull request 😃 
Could you please, instead of using a 'naked' async task, use the request pattern as used in the rest of the app (see for [the task folder](https://github.com/Awful/Awful.apk/tree/master/Awful.apk/src/main/java/com/ferg/awfulapp/task) for inspiration 😉 )

Also please refactor your code to still display the action menu and instead use the callback to just change the title.

Thanks 😃

Ephemerial:
> Also please refactor your code to still display the action menu and instead use the callback to just change the title.

@Sereri Not familiar with the code base, what callback are you talking about?

Ephemerial:
After a bit of fooling around, it does not seem possible to make use of the `AwfulRequest` class in this case. This is due (as far as I can tell) to JSoup throwing an exception when asked to parse an Imgur URL, and would likely require a refactor of the `AwfulRequest` to fix. I pushed these changes to the [awful-request-refactor](https://github.com/Ephemerial/Awful.apk/tree/awful-request-refactor) branch of my fork if you want to take a look.

Regarding the second half of your request: the title change must take place *before* the `PostActionsFragment` is shown. I could refactor the aforementioned class to expose the relevant view instance?

baka-kaba:
Sereri's saying it would be better to not 'block' the menu by waiting for the network request to finish before you display it. At best it makes the UI lag a little, at worst it could take a long time or even time out if there are network problems somewhere. (And there's only one AsyncTask thread so they end up queued - if there's a delay and the user keeps tapping on the image trying to get the menu up, there's a lot of waiting in their future!)

So it would be better to let the menu pop up, keep a reference to it somewhere, and have whatever async callback you use basically update the menu title when/if it gets the size data back. A WeakReference or something would probably be best, just in case the menu's been cancelled or the activity has been destroyed before the network response comes back

I'm not sure how easily it'll fit into the Request pattern since you want the actual connection - there's a ``RedirectTask`` that does some raw network connection stuff, and the callers extend it with the ``onPostExecute`` callback. But you have to be careful of those AsyncTask caveats (and definitely make it cancellable if the dialog is dismissed, so it can't block the thread). Honestly you might want to spin up your own temporary thread in the task

Ephemerial:
The action menu should now be shown immediately, and the image size displayed afterwards.

baka-kaba:
For your request refactor branch, I'm getting a lot of "Loading Failed" popups on images in [this thread](https://forums.somethingawful.com/showthread.php?threadid=3787824&userid=0&perpage=40&pagenumber=899). One of them works (the cartoon of a guy talking about his Control key or something), the others are failing though

But your Request is actually fetching the full image, and then you're opening a connection to check the size in the result callback. That's why I don't think the ``AwfulRequest`` stuff is the right place for something like this, it's tied a GET/POST system. You *could* try subclassing the actual Volley ``Request`` class, use something like ``Method.HEAD`` and try to get the content length out of that response

Or failing that, ``RedirectTask`` is probably more like what you want, a lightweight task that just opens a connection and does something with it, then closes it again. I still think it would be better to avoid an AsyncTask though, especially for networking - maybe a singleton class with its own private thread, so it can automatically cancel running tasks if it gets a new one? 🤔

baka-kaba:
The results in the master one are looking good though! I will use the hell out of this, clicking a gif and realising you just committed to a massive download is a bad time.

I have a suggestion actually - how about instead of rewriting the title, we have a separate ``TextView`` as a subheading and you write to that instead. So it would look like
# http://blablabla.bla
### Size: 1.2MB

something like that? Just make sure visibility isn't set to ``GONE`` so the UI doesn't jump when it gets updated

Honestly I think that title layout needs tweaking anyway, what do you think Sereri? The text is so big you almost always have to scroll it. Chrome's link menu has a really small title for the URL so it can fit a couple of long lines. I feel like that would look better, especially if we're adding a subheading for the size

Oh also, 'size' or 'file size' might be better than 'image size', that way it'll work for any link

Ephemerial:
@baka-kaba How does this look to you?



---

> That's why I don't think the `AwfulRequest` stuff is the right place for something like this, it's tied a GET/POST system.

I feel the same way; however I don't have very much experience with threading or the Awful code-base, but would definitely support a better alternative to using `AsyncTask` for this.

Sereri:
Yeah, sorry. You're right.

I like this version. Without checking, is the new textview hidden when displaying the action menu for links?

baka-kaba:
Looks good! You still might want to put 'Size:' or something in there - probably neater to keep it blank until you actually get a number to show.

``AwfulRequest`` is sort of a custom API for Volley's ``Request`` system - there's an inner class in there called ``ActualRequest`` which is the real subclass, it's fairly restricted in what it can do (like automatically doing either a GET or POST depending on whether there are headers). It's meant for making calls to the site, really. It's also a bit wild in there

I made a custom ``Request`` at 8c85bfe6709a58038a8bd0ed63f9cfc6454d6ab6 to give you the idea - it follows [the Android docs example](https://developer.android.com/training/volley/request-custom.html) and basically makes use of Volley's networking and threading stuff, so Volley manages the threading and delivers the result on the UI thread.

Just be careful though - you can't assume the same dialog/fragment will be displayed when you get the results back, so you can't just go setting the title without doing some checks. You could do a weak reference/null check in the listener, or you could use the request cancelling system: ``Request``s have a ``setTag(Object)`` method, and you can cancel all requests with a certain tag by calling ``NetworkUtils.cancelRequests(Object)``.

Volley [won't call the response handler on a cancelled Request](https://developer.android.com/training/volley/simple.html#cancel), so you could have a static tag for ``ImageSizeRequest``s, and then in any relevant lifecycle things (like if the dialog is destroyed) you could have it cancel any pending requests, and also cancel them before you add a new one (so there's only one running at a time)

Ephemerial:
@baka-kaba Thanks, I folded your example code into the fork.

@Sereri It is now. This does result in a small UI jump when the request finishes though.

Sereri:
@Ephemerial Something like [this](http://stackoverflow.com/a/17513664) might work. Or we could "prepare" the view if its an image and the request is performed by setting the visibility to invisible (or visible and empty string) before showing the view, then changing when it's loaded. That too would fix the jump.

baka-kaba:
So long as it's not visibly jumping (or moving targets around so people click the wrong thing) it's all good! Looks neat, should we change the URL text size while we're at it? Like maybe 1 or 2sp bigger than the size text

Sereri:
Having tested this for a few days now, I'm not the biggest fan of the background color changing. Maybe it would be better for the textview to always be visible (or gone when it isn't an image action view) and just set the text from empty to the filesize after it has loaded. Seems to work well however.

@Ephemerial It's up to you if you want to change that. If not, I'll merge the current state but might change it in the future.

Ephemerial:
@Sereri I'm not married to the animated fade in/out or anything, just figured it looked nice.

baka-kaba:
I finally actually merged in the most recent version, thanks Intellisort or whatever you're called!

I like the text fade in, that's nice, but yeah the background fading is distracting. I think it would be better to just have the title space for the TextView and have the text appear if/when the size data comes in. Maybe it would be easier to keep it visible and just animate the text opacity instead of the whole View

Looking good though!

baka-kaba:
I had a poke at the layout since we need to change the URL size anyway (9d261a3d99dcc0beae108b3ca55ba34dc5f9aa1a), I moved the header into its own LinearLayout so the background is in there and the text can just appear (also centralises some stuff and keeps the padding consistent if we end up making the size view disappear)

image

I basically copied Chrome's url style by eyeballing the text size, I changed it to max 3 lines too. I know it looks a bit weird but it depends on how the url happens to break across lines really. Good/bad?

baka-kaba:
Actually having used it a bit I realised it messes up the Post Options menu (which needs a normal title) - it's probably worth refactoring the menu dialog a bit, so you can just easily get a Post Options version or a URL options one without having to build and configure them (two ready-made recipes basically). I don't mind messing around with that, but how does the compact URL bit look?

Sereri:
>I don't mind messing around with that, but how does the compact URL bit look?

Looks good although the padding seems ... off.
Maybe try the look from @Ephemerial 's last screenshot: remove bottom-padding on the layout and have a smaller bottom-margin on the size text-view ?

baka-kaba:
I just followed the Material content guidelines specs for simple dialogs, so the whole thing is treated as a multi-line title with standard padding. If you prefer the previous look that's cool, where the padding goes just depends on how it gets refactored I guess!

Sereri:
I see. However I would like to treat the size more like an overlay to the actual header than part of it I guess. Especially if it only appears after a second (if at all) and the space it occupies is empty until it's loaded. Maybe it needs a text-placeholder or a spinner or something to occupy the space ?

baka-kaba:
Yeah there's always gonna be some empty space there while it loads (so long as we don't have the header expand, having menu items move seems like a bad idea), it's more how it's distributed and where the size bit ends up. 

Here's how it looks with the original spacing and the new font size

image

I think it would be best to have the size view optional, so it's set to ``gone`` for any menu except images? So this is like a special dialog layout and the other versions don't have to deal with it

Another possibility is to have the url centred in the header (which is a fixed size), and then if the size comes in, that can appear and the url can slide up to make room. That way there's no big empty space waiting for the size, but when it appears it's only the text that moves within the header, the rest of the menu stays exactly where it is. Calculating the size of the header in sp + separator dp might be awkward though

But yeah placeholder text might be best - that way you can add something like 'couldn't get size' if there's an error, and people aren't waiting around

baka-kaba:
Would it be better to just let this PR go through without the font size change? Maybe add the placeholder text with ``#setTitle``, so it all updates nice, and then deal with the smaller title text and the variation in dialogs with a later PR? I feel like it's bringing in these other issues and complicating Ephemerial's feature, and it would probably be better as a two-stage thing

Sereri:
Sounds good

Sereri:
@Ephemerial I'll add you to the acknowledgement list in the about section. Do you want your GitHub username or whatever your SA username is?

Ephemerial:
Cool, thank you very much! My SA username (Literal Hamster) works for me.

Sereri:
@baka-kaba https://developer.android.com/preview/features/autosizing-textview.html I guess this might be a good idea? Supported through the support library as well.

Issue #266, opened 2013-12-10
Sticky forums to the top


Sereri:
Similar to the functionality of Awful.app

baka-kaba:
Uh I think this can probably be closed now we have favourites? Unless there's some other functionality it's missing, like being able to reorder them (which is sort of a pain)

Sereri:
Nah, this is a good solution to this issue

PR #488, opened 2017-05-02
Sort Stickied Threads to Top

Ephemerial:
This wound up being a simple fix. Stuck threads will now be sorted to the top of the list in the thread list view.

Resolves issue #266.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
@@ -1425,7 +1425,7 @@ public class ThreadDisplayFragment extends AwfulFragment implements SwipyRefresh
 
 
     private class PostLoaderManager implements LoaderManager.LoaderCallbacks<Cursor> {
-        private final static String sortOrder = AwfulPost.POST_INDEX + " ASC";
+		private final static String sortOrder = AwfulThread.STICKY + " " + AwfulPost.POST_INDEX + " ASC";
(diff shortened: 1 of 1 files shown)


Sereri:
Hi Ephemerial,

unfortunately the ticket in question (#266) is defined rather vaguely, that's my bad. The proposed feature actually suggests adding some kind of toggle to the items of the forum index (the left-most view) to allow these forums or subforums to be displayed on top of the forums list (below the bookmarks) as well.

The choice not to "honor" stickies in the bookmarks is a deliberate one on the other hand. While stickies are used to single out threads deemed 'special' by the forum mods and admins, the bookmarks contain threads deemed important to follow by the user. In the bookmarks, stickied threads are therefore among equals. As such, the user preference supersedes the sticky in this case and stickies should not float to the top.

I hope this will not discourage you to submit further pull requests in the future 😃

Ephemerial:
@Sereri My testing does not show stuck bookmarks being sorted to the top.


---

Ephemerial:
Testing the production version I use on my phone seems to indicate that stuck threads were being properly sorted to the top of the forum thread list all along. I could have sworn that this behavior was absent; maybe I was thinking of an earlier version?

Me dumb, closing PR.

baka-kaba:
No you're right, I just clicked on the Games forum and the top threads are sticky, normal, normal, normal, sticky, sticky, sticky. Bookmarks having no particular order is fine, but we want stickies at the top of the regular forum view, right? (And we can *maybe* stick them all in an expandable dropdown at some point)

Sereri:
Funnily enough, they're _sticky sticky sticky sticky normal..._ for me. This might need investigating after all.

PR #495, opened 2017-05-15
Add long-press title popups on images, update Zepto to 1.2.0

baka-kaba:
I've added that request for smiley code popups, it basically just tells the app to pop up a toast with the contents of the ``title`` attribute. I had to upgrade Zepto to get the touch stuff working, I rearranged things in the assets folder too


baka-kaba:
I can't test it on my emulator, the popup works fine but the tap-to-play thing doesn't - it didn't before I made the changes either. Any chance you can check it on an old device Sereri?

Sereri:
It works on my 4.0.3 device as well as on my 4.4 device. It just doesn't work very well sometimes. I think a doubletap would've worked better for it but this doesn't work because we're using tap (or click doesn't matter) for the other event. . It looks awful on the ICS device but, well, who cares about those people.

baka-kaba:
Cool, thanks!

Zepto has ``singleTap`` and ``doubleTap`` events so we could always try it, I just thought the long press was more discoverable since it's the Android Way

What problems are you seeing? The long press selecting nearby stuff? I get haptic feedback before I've pressed long enough to trigger it (the minimum press is 750ms) which is awkward but it always works for me so far

Sereri:
Oh, I only saw the normal tap, which replaced the element causing the doubleTap to never work. I'll try this when I get home.

As for the problems, the longtap sometimes selected the emoticon, sometimes text next to it and sometimes nothing happened at all. It's working ok on my Marshmellow N5 though.

baka-kaba:
When it selects text, does the long press event fail? Like I said the delay's a bit longer so you have to keep pressing after the selection triggers (looks like you can tweak the delay constant easily in ``touch.js`` though)

Sereri:
Yeah it does.

I tried it with singleTap and doubleTap and it works quite well, except when it doesn't. Some of the time it will trigger doubleTap even if the second tap is a second after the first.
I'd merge it in as-is I guess.

baka-kaba:
Yeah it'll do, we can always pull it if there are too many issues

PR #498, opened 2017-05-25
Refactor post menus and add small URL titles

baka-kaba:
I've refactored the post menus to hopefully make them easier to maintain, each menu has its own subclass and all the definition and action handling happens in there. Some of the called code has been simplified, e.g. to handle integer IDs as integers instead of Strings, performing the conversion when it's actually needed

I've also tweaked the layouts a bit, to make them fit material guidelines - margins and padding align with the keylines and the icons are smaller. The URL link menu has smaller text and has an optional subheading view (only used for images currently)


baka-kaba:
Ok I had a crack at this, it's not done but how does it look in general? I tried to make something that's easily tweakable or reusable for other menus. I used enums because it looks neat, but you can provide the actions however you like

I don't really like that there's two layouts and they share so much xml, I like being able to see changes in the layout designer but it might be worth just using the URL one (with optional subheader) and defining the title style programmatically instead. Or defining a ``url_title`` style and just applying that, that kind of thing

I messed with the actual layout design a bit to make it more ~material~ so there's [smaller icons](https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-touch-target-size), [keyline alignment](https://material.io/guidelines/components/dialogs.html#dialogs-specs) etc. It looks... different but maybe the inverse title bar is throwing it off too.

I also added a ``min-width`` to the layouts to force the dialogs to meet the ``280dp`` spec, otherwise they resize to fit the contents - which definitely looks neater when it wraps the labels nicely, but it means the menu dialogs look really inconsistent depending on what you clicked on, some are super-narrow. The wasted space might be better but what do you reckon?

I also moved some padding where it 'should' be (in the container) but that messes up the drag bloop thing on the recyclerview so that needs fixing I guess

Sereri:
I think it looks nice. I'm not sure why this looks better to me than the last time you proposed smaller icons. Do you remember what's the difference between this and the previous one?

I'm not sure about loading file-sizes for all images instead of just for gifs. Most of the time when you tap on it, the image has already loaded and it's basically only telling "you downloaded 100kb GJ"

I also have some bug where it doesn't show the animation and also either doesn't show the size or shows it at like 0.5 opacity

baka-kaba:
Uhh not sure about the small icons thing! I probably didn't adjust the margins or whatever, this time I basically made sure everything meets the specs. Leave it to the design pros I guess is the lesson here. The minimum dialog width might be helping too, sometimes it looked too weird when all the content was narrow

Good call with the gifs-only thing, I was sort of thinking it could be there for all links (in case you want to download something) but then I though who cares, nobody's linking pdfs on the regular or anything

That bug is weird though, I've been using it for a while and I've never seen that. Did you do anything special? Or does it stay weird and broken after the first time you see it? Or is it just completely random?

Sereri:
I'm not sure, I _think_ it worked once and I thought neat, then it stopped working and went randomly opaque/invisible

baka-kaba:
Ok yeah, the view wasn't set to GONE like it should be so there was always a fade in with the initial text, then the update came in and assumed the mid-fade alpha was the target and it ended up stuck at that. I guess I had seen it, it was just close enough to 100% that I couldn't tell 🦅 

I took out the save-and-restore thing anyway, the fade doesn't affect text colour alpha and that's the only thing anyone should be changing in XML

baka-kaba:
So is this one ok to merge in when I tidy it up, or does it need anything else?

Sereri:
Sorry, I was away on a workshop this week. I just updated to the new version an it seems fine. Feel free to merge.

baka-kaba:
No probs, I'll probably end up refactoring a little first

PR #501, opened 2017-05-27
Add favourite forums

baka-kaba:
I know, they're all so good how could you even choose right?!

So this basically adds a Favourites star to the menu for the forum list, which allows you to switch between the full forum list and the favourites view. In favourites there's another icon to actually select your favourites from a list, which happens through a popup dialog. There's also an option in the settings to default to the favourites view instead of the full list.

Favourites are stored in the sharedpreferences as a list of forum IDs, so they should survive backups etc. The forums repo lets you pull out those forums from the database, so you can display them as a list or work out which items to check as current favourites in the forum list.

Currently the favourites are just presented in the same order as the full list, you can't change their positions. It wouldn't be too hard to implement in the repo (just get the forums from the DB and reorder the list to match the stored IDs), it's just a lot of work creating the UI to let the user move them around. Which is why I went for the checklist too - a contextual action mode on the full forum list would be nice, but it looks like a headache to implement as things are


baka-kaba:
This is all done and tested and ready to squash and merge, just wondering if there were any suggestions

Like I said I've sort of taken the easy route with the checklist, but it's a lot less complicated (and discoverable) than trying to implement a contextual action mode. That could come later I guess, it's not too bad as it is really though, and it's not something that you need to interact with much

Sereri:
I'm not sure about the implementation. Like, I get what you did and why, but it feels weird wrong somehow. 
Is it ok if I add it to the beta first without merging it in and we'll see what the feedback is like?

The only thing I actively dislike though is the icon for the forum selection. I have no idea what it means or what it's meant to represent. I think even a settings cog would be better.

baka-kaba:
It's a screen with boxes on it or something! Jeez!

Nah I was just trying out some new icons, a list with a plus (or just a plus) would be better I think. A cog could work but I think that implies more than just settings, but whatever works for people. Feel free to beta it if you like - what aren't you sure about though? The checklist, or the way it uses menu icons, or the whole thing?

The checklist could go in the settings but honestly, even though it's not something people will change much, I wanted it to be discoverable. Right now there's a mysterious star, and when they press it the Favourites view pops up saying you have no favourites, but a new icon has appeared - so ideally people will be led through the whole UI and learn how to use it within a few seconds. I guess the 'no favourites' label on the empty list could say 'go in the settings to add some' but that's not as clean, plus nobody ever looks in the settings anyway

But yeah the checklist thing isn't ideal, but honestly I don't know if I'd have time to do anything more complicated. Although maybe I can bang out a long-press add/remove thing

Sereri:
I don't know, the fact that it's a second screen feels weird (though it might be the best idea), instead of putting it between bookmarks and Gibbis.

> when they press it the Favourites view pops up saying you have no favourites, but a new icon has appeared - so ideally people will be led through the whole UI and learn how to use it within a few seconds.

This is where I'm also a bit cautious, I think a few people might go "oh no the forums are gone, I guess this is my life now". Actually, thinking about it, an empty-text might go a long way here. Just "Press {Icon} to add forums to your favorites" or something like that

baka-kaba:
Well the idea was to have a switchable view of the forums, either the full list or just your customised list of favourites. You can set favourites as your default view if you want, but that's off by default, so if anyone did get confused they could just restart the app and it will be back at the full forums view. Plus the title changes so it should be clear that the icon they pressed just switched things over

I'll have a look at getting a message with the icon in there, at the moment I'm just repurposing the 'no forums data' view. I changed the icon too, is this one better? 

I added a context menu to the forum items so you can toggle the favourite status - it's unfinished, but does that feel better? It's sort of awkward because it'll involve some messing around to get items to disappear from the favourites list when you unfavourite them, plus you'll probably need some kind of visual indicator of which forums are already favourited on the main list (a star like the sticky icon on threads?) , and there'll need to be a callback to update that whenever the status is changed

For something people will only change rarely it feels like it's overcomplicating things - the checklist is clear and simple (IMO anyway) and all the favourites stuff is kept separate in that one customisable view

PR #509, opened 2017-06-28
Add Imgur uploader/inserter

baka-kaba:
Shows a dialog where the user can select an image / add a URL and have
it hosted on Imgur, automatically inserting the hosted image as BBcode
in the message composer.

The app needs an API key for Imgur - I've added a secrets.xml file to
the .gitignore to add the key as a resource that's not checked in. The
free tier puts credit limits on the app and blocks it if too many
requests are made, so the ImgurUploadRequest class keeps track of the
current limits and enforces a per-user daily quota (to stop a few users
using all the app's credits)

The credit limits might need tweaking for a better experience, depending
on how many people actually use it and how much. I just guessed at some
limits!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ tests
 out
 build/
 gradle/
+
+# Secrets as resources
+**/*/res/values/secrets.xml
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -110,4 +110,5 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.hamcrest:hamcrest-library:1.3'
+    compile 'com.android.support.constraint:constraint-layout:1.0.2'
 }
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/reply/BasicTextInserter.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/reply/BasicTextInserter.java
@@ -17,7 +17,7 @@ import static com.ferg.awfulapp.reply.BasicTextInserter.BbCodeTag.FIXED;
  * Handles inserting basic, unparameterised BBcode tags into EditTexts.
  */
 
-public abstract class BasicTextInserter extends Inserter {
+abstract class BasicTextInserter extends Inserter {
 
     /**
      * Wrap selected text in a BBcode tag, or show a dialog to insert some.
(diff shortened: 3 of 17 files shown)


baka-kaba:
That's right I did it

I've been using this a while and it seems to work pretty well. The whole credits system is sort of a mess and I've had to guess at some useful limits to stop goons from spending all the credits, maybe @nolanw has some suggestions - right now it's 50 uploads/500 credits per day (1,250 uploads max for the app, I've cut it off at 1,150 though so we don't get _**banned**_

The UI layout is... not great, I was playing with ConstraintLayout but I haven't really found a nice-looking way of arranging things, URL mode especially looks kinda crap and busy. I mean it works fine, could just look a bit better. Open to suggestions on that for definite

Anyhow the app needs [registering for a (free) API key](https://api.imgur.com/oauth2/addclient), I've set up a secrets.xml resource file in the gitignore, I don't know if we're using another secret-handling thing for the app?

nolanw:
The free Imgur API thoroughly baffles me. I'm afraid I don't have any useful suggestions.

baka-kaba:
Wait you're not paying for goon uploads are you? Or are you using some library that takes care of all the details?

nolanw:
We're just hitting the API and failing when it fails. Not bothering to check limits or throttle or whatever.

And nope, not paying.

baka-kaba:
Well that's a relief

Yeah the credits are really restrictive, they rate-limit users but only to 50 uploads per hour, a few people with lots of cool memes they just have to share could burn through the daily quota pretty quick. And if you hit the limit 5 times in a month the app's blocked for that month. Does it usually work out ok for you guys? Can you usually upload something when you go to use it?

nolanw:
I can, but we get perennial reports of users getting errors as the month goes on.

I guess it works ok, I don't really think about it. Maybe everyone got fed up and uploads images separately.

I do have a todo item to allow logging in to Imgur, which I think has a different rate limiting scheme.

baka-kaba:
This can go into the dev build I think? It's basically done (might add a fullscreen dialog for small screens later), just needs the API key location deciding unless it's fine in the secrets.xml file

Sereri:
Yeah sure. The only thing still that might need adjusting is the limit. We have 10k users and 1k uploads so it's probably a good idea to bring that down to 1 or 2 during the start where everyone will try it.

If the tokens are spent and the user has the imgur app installed maybe we could launch an intent at it instead?

baka-kaba:
Well I was thinking the beta would be a good place to fine-tune the upload limits, just to make sure we're successfully avoiding the ban limit. And yeah, restricting the quotas so more people get a chance to upload a thing and go 'oh cool' is probably a good idea. I doubt all 10K users are gonna use it, but I'm sure some will be going wild. Maybe do that before the normal release though? Just so the beta can shake any bugs out with 'normal' use

The intent thing is a good idea, I'd want to look at that later though - I'd rather just get this done and out the door

PR #525, opened 2017-08-19
Add Instagram and Twitch embedding, add separate JS file

baka-kaba:
I've added embedding functions for Instagram and Twitch URLs, and preferences
for both. Instagram defaults to on (for discovery), Twitch defaults to off because
it is *really heavy* and a page with multiple videos has to load some embedded
player for each link. Showing a preview pic that's play-on-click would be better,
but that requires an API key...

I added these functions to a separate script file where all the embedding code
can go, just to make it easier to extend without complicating the general thread
setup code. So now there's just a single call to the embed script, and that runs
everything according to the preferences that are passed in.

Also added some Zepto libraries to enable Promises - handy for things like the
Instagram oEmbed code, where you need to add HTML for each element
and then process the finished page when they're all done.

### DIFF
--- a/Awful.apk/src/main/assets/javascript/embedding.js
+++ b/Awful.apk/src/main/assets/javascript/embedding.js
@@ -1,0 +1,40 @@
+//
+// Created by baka kaba on 16/08/2017.
+//
+// Functions to automatically embed page content, e.g. turn an Instagram URL into a widget
+//
+
+function processThreadEmbeds() {
+
+    // map preference keys to their corresponding embed functions
+    var embedFunctions = {
+        "inlineInstagram": embedInstagram
+    }
+
+    // check all embed preference keys - any set to true, run their embed function
+    for (embedType in embedFunctions) {
+        if (listener.getPreference(embedType) == "true") {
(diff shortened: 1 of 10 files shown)


baka-kaba:
I think it's worth doing this separate embed file and putting all the fancy content stuff in there, just so it's easier to work with and doesn't fill up ``thread.js`` with its basic setup stuff. There's an issue about embedding at #526

The Instagram embed works pretty well, behaves itself and looks nice. The Twitch one is sheer hell if you just let it run on every link, it loads this monster player widget and takes forever. If it's worth including we might have to look into making it click-to-run, but seems like you need an API key if you want to get a thumbnail easily...

Sereri:
So, uh, any reason why you didn't merge this in? Looks fine to me.

baka-kaba:
I wanted to leave it while I poked around at the Twitch stuff, and I'm not sure if it's worth adding that in as it is - you get a thread with a few of them and it takes ages for the page to load. I left it open to see if anyone had any suggestions, really

The Instagram stuff works nice though, they have a good widget

PR #537, opened 2017-10-29
Fix https errors

baka-kaba:
All right, I found a solution which is basically using [this Google library](https://developer.android.com/training/articles/security-gms-provider.html) to patch the secure networking provider. One thing it does is stops older devices from insisting on trying SSLv3, which I'm guessing Cloudflare just disabled

I tried some of the other stuff with SocketFactories and whatever, didn't work - still went for that SSLv3 even when you remove it from the enabled protocols. This seems to work ok? Tested on an API 18 (4.3) emulator, fixes the problem - not sure about earlier APIs, can't test those

One problem is avatars, emotes and some special icons like admin stars aren't working on the emulator. Fine on my phone though, hopefully it's just a problem for old devices so it's either no pics or no connection at all. It's just a quick fix anyway

### DIFF
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -91,6 +91,9 @@ dependencies {
 
     compile 'com.android.support:appcompat-v7:26.1.0'
     compile 'com.android.support:design:26.1.0'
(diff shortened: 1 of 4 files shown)


baka-kaba:
Had to redo this for master (it's so old!), so I applied a hotfix to that instead

PR #547, opened 2017-12-05
Rework and optimise post parsing

baka-kaba:
I've redone the code for parsing posts from a thread page, to make page loads
faster and hopefully easier to maintain. Parsing is multithreaded now, it's more
specific about Element selection (to avoid expensive tree parsing) and uses some
jsoup improvements. Overall I'm seeing 3x faster parsing for the post elements,
which cuts page load times by almost a second on a full 40-post page.

Feels a bit snappier, and hopefully it'll get faster when jsoup updates to handle
Windows-1252 encoding in an optimised way. Ideally we can improve other parsing
tasks too (maybe using a shared executor for all parse tasks) and make loads as
quick as possible - they're pretty slow.

I tweaked AwfulError a little too for some speedup (since it's run on every
request response), but there's only so much I can do without being able to see
the content of error pages. I've also removed some code from NetworkUtils which
shouldn't be needed anymore (the anti-logout stuff that should have been fixed
by limiting the number of pending requests)

### DIFF
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -120,7 +120,8 @@ dependencies {
     //compile 'com.mcxiaoke.volley:library:1.0.19'
     compile 'com.github.samkirton:android-volley:9aba4f5f86'
     compile 'com.google.code.gson:gson:2.8.0'
-    compile 'org.jsoup:jsoup:1.10.3'
+
+    compile 'org.jsoup:jsoup:1.11.2'
     compile 'com.jakewharton.threetenabp:threetenabp:1.0.4'
     compile 'com.samskivert:jmustache:1.13'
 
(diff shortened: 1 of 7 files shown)


Sereri:
I totally forgot to comment on this. I think this is pretty cool and I'm surprised that (even with threading) it's possible to improve parsing by a factor of 3. Good work

baka-kaba:
Hey thanks! Mostly it's using the new ``selectFirst`` stuff that returns when it finds an element instead of walking the rest of the tree, it ain't fast. I could have sped it up more by being really specific about where things are, but the way Lowtax keeps messing with the CSS I'm wary about tying it too closely to a particular structure

I was worried it was just my slow phone that would see any difference but hey, anything that makes it feel snappier is good right!

It'll get a fair bit faster when jsoup adds that fast path they have for utf-8 and ASCII. I just wish network requests were quicker now (again might just be me)

PR #553, opened 2017-12-20
Updated and alphabetically sorted contributers list

Sereri:


Sereri:
I totally forget I did this

PR #559, opened 2017-12-31
Add Recent tab to emote picker, improve filtering, add layout tweaks

baka-kaba:
I've reworked the EmoteFragment class into a Kotlin file - now there's a tabbed
ViewPager with the normal search view, and a Recent view that shows the last
emotes you've used. The fragments have default constructors and don't hold any
references to listener components, so they should be better behaved if they
need to be recreated.

The Recent view shows the last 30 emotes used - this is basically a queue that
drops the least recent emote when it fills up. It doesn't take frequency into
account or anything. There's no favourites handling either, it could be done in
a similar way but the context menus are awkward. Maybe later!

Filtering now works with multiple terms, separated by spaces - so you'll get
results that match any of them.

I've reworked the layouts a bit while I'm at it - the clear filter button is
Material-style now, I've added some padding to the dialog, and the emote labels
shrink a little if their text is too big.


baka-kaba:
Any suggestions before I merge this in?

Awful.apk/src/main/res/layout/test_fragment.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Sereri: Did you mean to commit this?

Awful.apk/src/main/java/com/ferg/awfulapp/EmoteFragment.kt
```diff
@@ -0,0 +1,284 @@
...
+            }
+        }
+    }
+}
+
+
+/**
+ * Class with shared code for the fragment pages, mostly handling the emote grid and lifecycle stuff.
+ */
+abstract class EmoteGridHaver : AwfulFragment() {
```
Sereri: Maybe Owner instead of Haver? Or maybe something else that isn't Haver :V

Sereri:
Oh god how did this get in here I am not good with computers

Also I need to sit my ass down and finally dive into kotlin

Awful.apk/src/main/java/com/ferg/awfulapp/EmoteFragment.kt
baka-kaba: All right, you change-haver!

Awful.apk/src/main/res/layout/test_fragment.xml
baka-kaba: Heh like I need to make a test fragment, heh heh *tugs collar*

Sereri:
I've no issues with the rest of it although it might have to do with not being sure whats going on in half of it.

Do you think that filtering could be improved by parsing and looking through the emoticon titles as well?

baka-kaba:
Messing with Awful is a good way to learn it I reckon. Go hog wild and get the bad habits out of your system...

Is it too weird though? I worry I'm going a bit too... functional and turning bits of the project into something confusing

Sereri:
Looks good 😃

baka-kaba:
Just rebased and added the proper ripple background on the delete button

PR #560, opened 2018-01-01
Kotlin conversions, add Timber logging library

jsulli:
Heavy kotlin conversions. Converted most of the main activities and fragments. 

Functionally, I made barely any changes, I wanted to do a straightforward conversion to avoid breaking too many other things in the process. My main goal is to get the bulk of the main classes in kotlin before I start ripping other things apart. Most changes were around making as many things immutable or non-null as possible and adding a little syntatic sugar with `let` `apply` `when` etc. It's about 15% LoC reduction right now, more thorough conversion could get that to around 30% pretty easily. 

I also added Timber, and changed all the `Log` calls in the classes I converted to it. I set up a basic Fabric tree for release builds.

There may be lingering nullability issues but from my testing pretty much everything should be good. There is a crash when attempting to preview a post - but prior to my changes this operation failed on the dev branch anyway, I plan to look into that next and get it working properly and do another PR.


Sereri:
I'm not sure if I'd not rather have had 20 different PRs 

This might take a while

PR #566, opened 2018-01-06
 Refactored thread.js

Sereri:
- Attached event listeners to container where possible and moved adding them to container creation.
- Renamed a few variables
- Moved some anonymous functions out into named functions
- Put some jquery requests into variables
- Reworked the madness that was signature hiding.
- Added comments


baka-kaba:
Ahh javascript
nodnod
Yeah that looks better. While you're in there, is there an easy way to cancel duplicate pending events? Like if a heavy page is loading, and you tap an image link or the post menu, nothing will happen - so you tap again, and when it finally catches up, it creates two menu dialogs because there were two "show menu" events queued up

I'm guessing no and we have to fix it in Java, but it's worth asking!

PR #570, opened 2018-01-08
Page bar upgrade

baka-kaba:
- Reimplemented using ConstraintLayout
- Tweaked icon and font sizes, added minimum width to page number display to keep the overall spacing stable
- Resized touch targets to Material spec (accidentally touching the edge of the screen and going back a page was driving me crazy)
- Added button ripples to all elements, and included a couple of Material ImageButton styles to use generally
- Added layout change animations, reworked PageBar class to handle display states better and show/hide elements in the correct order to get clean transition animations


baka-kaba:
I'm pretty sure I asked about this a while back, but could we reduce the icon and text sizes on the page bar (like I've done here)? It all looks cartoonishly huge to me, and seeing the nice neat page counter on the iOS version made me want to try refining it.

I dropped the icons down to normal material spec size (24dp) with the 48dp target, and set the text to Body2. I think with the borderless ripples and animations the whole thing just feels slicker without being harder to use (it's actually easier since you can't misclick in the empty space now)

I'm not opposed to a slightly larger text size either, but here's how it looks on my Pixel 2 emulator at default text size:
image

Here's a few font-size comparisons
Smallest:
image
Default:
image
Largest:
image


And this is the original
image

Smallest:
image
Default:
image
Largest:
image

I think those are the same (it's the ActionBar text style) but uh I'll leave them there for posterity

baka-kaba:
Also how do we feel about dropping the word "Page"? I guess we have a few options
- _Page 9/10_
- _page 9 of 10_
- _9/10_
- _9 of 10_

Dropping the word entirely looks nice but it might be a bit ambiguous (you'd definitely want it there when we don't have a page count, so like _page 1_ instead of just _1_). I think it might look better lowercase, and maybe _of_ would look nicer than the slash

I can add some quick mockups while I'm at it

Sereri:
@nolanw and the Awful.app people uses the **9 / 10** variant, I'm sure they have carefully thought out reasons for that. :v:

Generally I'm a bit ... concerned(?) about the ~~white~~blackspace this will cause, I do however like that the font-size is relevant now

Awful.apk/src/main/java/com/ferg/awfulapp/widget/PageBar.java
```diff
@@ -86,30 +86,54 @@ private void init() {
...
+        } else {
+            type = PageType.ONE_OF_MANY;
+        }
+        // if currentPage is greater than lastPage, then lastPage isn't a meaningful page count (-1 is passed in when we don't have that data anyway)
+        boolean hasPageCount = lastPage >= currentPage;
+        updateDisplay(currentPage, lastPage, type, hasPageCount);
+    }
+
+    private void updateDisplay(int currentPage, int lastPage, PageType pageType, boolean hasPageCount) {
+        String text = String.format(Locale.getDefault(), "Page %d%s", currentPage, hasPageCount ? "/" + lastPage : "");
```
Sereri: I hope the irony of using string concatenation inside a String.format is not lost on you :V

baka-kaba:
You mean you're worried about the space if we used **9/10**, or the space as it is now with the smaller icons and text?

If it's the latter, we can always add some padding/bias to the icon areas to push them closer to the middle, so it lines up more with the phone's navigation buttons and feels less spread out. If it's with the **9/10** then that ``minWidth`` value will be coming down anyway, so there wouldn't be as much space

How do you feel about the visual change in general though? I've been using it a bit and I think it looks sleeker and a lot less intrusive, it sort of disappears when you're not looking at it. I might add a page count style randomiser just to spend some time looking at the different options

Awful.apk/src/main/java/com/ferg/awfulapp/widget/PageBar.java
baka-kaba: It's my boy ternary operator

Sereri:
Both more or less, although the 9 / 10 would probably exacerbate the effect. I'm also slightly sperging out about the arrows not being centered correctly although that looks to be the same in the screenshots of the current version

baka-kaba:
The arrows are centred in the space, it's just that minimum width on the counter I was talking about - because there's no visual guide to where that view ends, the space looks part of the arrow area. (It's the same on the original version, for what it's worth)
It also looks weirder on this randomising version because I picked a ``minWidth`` that fits "Page 9999/9999", so that's way too much if it's just the "9999/9999" format. Obviously you'd tweak it for the final format

I guess there's four options
- drop the minimum width and padding on the page number view - the icons always split the space equally, but they move around as the textview changes size (I think this would look pretty bad)
- pad the sides with ``minWidth/2`` (but if you ever change it you need to do the sides too, won't happen)
- split the bar into thirds
- don't try to centre the icons at all, and just move them closer to the page number bit

Could probably try getting them to line up with the Android nav buttons if you want, but I feel like that's a whack-a-mole scenario

jsulli:
I think it would make more sense to not constrain all these buttons to each other but just to the parent, that would eliminate the centering problem. Arrows would be constrained to right or left sides of the screen with margins. They don't need a constraint towards the center, if that makes sense. If you're worried about that allowing too much stretch on a wide display, there's an easy fix for that too:

Add an invisible view (that sits on the same level as everything else) and give it a max width while constraining it to the parent. Then constrain everything to the edges of that instead of the parent - then when it's on a tablet or landscape everything will still be grouped to the center while still being stuck to the edges.

Similarly I think the refresh button can be constrained only to the parent, with a 50% bias towards one side. You'll have to see how that fits next to a thread with 7000 pages or something but it looks like it'll fit fine.

While you're at it, there's a small thing that bugs me - when you're on the last page of a thread, the refresh button replaces the next page button on the right side - but when you're not on the last page, the refresh button appears on the left side between the _previous_ page button and the counter. Seems like that should just be moved to be always on the right. Either next to or instead of the next page button.

baka-kaba:
The centring thing is more of a perception - right now the page counter is in the middle, and the buttons occupy the space on either side, distributing themselves evenly through it. So if there's one button it's centred in that space, if there's two they divide it up so they're equally spaced out. (The original version has padding, so the gap between two buttons is 2x the gap on their other sides).

The problem is that space for the icons needs to be fixed to stop them shifting around - and the page counter needs a bit of room to show bigger numbers, so a lot of the time there's empty space there. It just doesn't look like it 'belongs' to the counter, so it feels like a larger area that the buttons aren't properly arranged in. It's not really a layout issue, it's more of a design one - how do you make a fixed element look centred in an area that perceptually shrinks and grows on one side?

So yeah I think it either comes down to a) set a close min width for the page number bit and don't worry too much about it, or b) don't centre them. I tried moving them in a bit closer to the middle and... I'm not a big fan, but we'll see (you can use guidelines to set those constraints you're talking about too). Or another thing is adding a subtle edge to the left and right of the page number area, so there's a visual hint that makes the centring work

baka-kaba:
As for the actual buttons... the rules are basically
- the refresh button is always visible and always next to the page counter
- there should never be 2 buttons on one side and 0 on the other
- the refresh button goes on the left, unless that breaks the previous rule
So the refresh button moves around to fill space if necessary, like when an arrow is missing. There used to be a refresh button always visible on both sides but that looked really weird to me

If you're saying it should always prefer the right side instead of the left, you might be right! That's probably more usable. And if you're saying the arrows should always be on the inside, next to the page counter... yeah maybe, that probably looks better but it might feel strange (tapping near the middle instead of the sides of the screen). I'll definitely try some stuff out

jsulli:
Sorry if I wasn't clear - I do think the arrows should be on the outside, I just don't think they should be affected by the movements of the other buttons. I think it would be more consistent and reliable, and would always be properly spaced. So they'd be locked to the edge of the view (with margins), but have no direct relationships with other buttons.

The refresh button IMO, should be affected by the placement of other buttons (i.e. it moves when there's no next page button), but other buttons should not be affected by it. I think there will be space for that. But these are just nitpicky design issues so it's up to you!

jsulli:
This layout has only one refresh button, and arrows that are only constrained to the edge. They will never move in response to the page counter or the refresh button. The refresh button is tied to the next page button, and will move automatically into its position if the next page button is set to `gone`.

Pictures! This is with/without next page button on the nexus 4 emulator and pixel 2 XL emulator, so a good phone size spread.
nex4-1
nex4-2
pix2-1
pix2-2

Layout: 

https://gist.github.com/jsulli/c52fe0f3fbf726b42932370aab5d9fd2

baka-kaba:
Oh ok, I see what you mean now!

Yeah honestly, it bothers me because you have the page number and arrows in visually specific positions, but then the refresh button looks like it's just _there_, it feels unbalanced - even moreso if it's only ever on the right (so first pages have nothing on the left and two buttons on the right)

That's just a problem with having to put that spare element somewhere - the idea with evenly distributing the elements on each side is to make it look more deliberate, so there's a balance to it no matter how many items are there, and the animation plays into that. It's definitely better to have things fixed in terms of usability, but they don't move that much and I think it's worth it for a nicer layout (all subjective obviously!)

I tried the thing with adding a visual guide to outline the spaces
image
Haven't spent much time with it, but I think that looks pretty ok? Maybe smaller, maybe a touch darker, but the buttons definitely look more _right_ when you can see the space they own

jsulli:
I really like those bars. And I definitely like the plain "/" or "of" instead "page". 

The slight uneveness to the layout is probably not a big deal - i've used this app for god knows how many hours and I never actually noticed until seeing these pictures, heh. 

A somewhat more far out option.... maybe we don't even need the refresh button? PTR is in place on both the top and bottom of a page, and it's such a standard feature in nearly every app now that I don't think anyone would miss the button. But removing features is always risky, so who knows.

Another option is to lock all prev/next/refresh buttons in place, and instead of hiding them, disable them, with the button turning gray and doing nothing on click. That's another pretty standard feature people will probably intuitively understand.

baka-kaba:
Yeah I did think of the dimming thing... but that doesn't really solve the refresh button problem, it either has to be on one side (although now you have a disabled left arrow which is a little better), or you have one on each side, like the old style (double the refreshment!) only sometimes one works and sometimes the other does, and it doesn't really reflect the space or anything, it's just some seemingly arbitrary limitation that you can't use the left refresh button if the left arrow is lit up

I've been thinking about thos page bars far too much obviously

Oh as far as pull-to-refresh goes, it's only an optional feature and it's not very accessibility-friendly - you're meant to stick a refresh option in the menu instead though. But you just know people would get real mad about taking it away

And yeah I like the _page_-less stuff too (lowercase _page_ looks better but it's still a bit busy). If you haven't tried this PR, try merging it into your daily driver build - it changes the style every so often so you can see a few variations, or add yr own!

baka-kaba:
All right, I've used this a bit now and personally I like this style best
image
with the numbers separated by a slash (with spaces, looks way better when the numbers are small like ``1/2`` which looks like a fraction). I like *1 of 2* as well but it feels... I dunno, fancy, and doesn't really fit the style of the app. Same for italics.

There's no "page" *except*  when the page total is missing - so it either says ``1 / 2`` or ``page 1`` (lowercase). That's inconsistent but it's clearer about what that single number is, but I'm open to just leaving it as the number and assuming people are smart enough to work it out.

I think the dividers work well too - you lose that nice clean field, but you gain that sense of space that makes it all look balanced, instead of icons looking like they're not lined up right. So I think it looks a lot cleaner as a result. I went for a darker colour so it's more of a subtle hint than a visual flourish or whatever, but I'm fine with darker or lighter too

Anyone got any thoughts or can I merge this in?

Sereri:
Looks good, feel free to merge it in if you're done

baka-kaba:
I've dropped the 'page' prefix for now, it always flashes up as a thread page loads, before the count comes in

PR #573, opened 2018-01-10
Twitter image fix

Sereri:
I'm bad at git.

 Resolve twitter images no longer loading by updating the twitterwidget.js, using a newer api and enabling the DomStorage of the webview


baka-kaba:
Looks good!

I still think it would be better to load the Twitter script when the document is set up, instead of having a local copy that might end up invalid when they change things, but yeah there's probably always a race condition between loading that and running it on the tweets. I don't know enough about JavaScript to suggest the best way to handle it though

Speaking of, I noticed that all the embedding code is in ``salr.js``, but I created a separate ``embedding.js`` file for that stuff, so it's probably worth moving one to the other so they're all together. I don't mind getting rid of the embedding file (I think the preference->function list is a bit cleaner than a bunch of ``if``s though)

Sereri:
I will move the embedding code from salr.js to your embedding.js during the vanilla.js refactoring

PR #576, opened 2018-01-13
Convert ForumsIndexActivity to kotlin

jsulli:
Significantly reduce and separate fullscreen functionality


baka-kaba:
Uh guess I can't just put this in a general review?

I'm seeing a few bugs here
- in non-immersion mode, the activity's action bar is below the status bar and the page bar is missing (I'm guessing it's below the nav buttons, it looks like everything's still laying out like it's fullscreen)
- hamburger button / back arrow doesn't work
- activity title isn't working (you deleted the call to ``#setActionBar`` in ``#onCreate``, it uses a custom view)
- the first time you tap a thread to open it, the app pauses and then nothing happens (doesn't page over). Works fine after that.
If it's something like creating the WebView fragment on demand, the current version does it in advance (even though it looks like it doesn't - it's a side effect of some other code that runs, this is why I said refactoring the view pager ended up complicated!)
- if you've opened a thread, the app crashes when the activity is restored after being destroyed (e.g. turn Don't Keep Activities on in Developer Options, go from the app to the home screen and back to the app)
- if the app isn't already open, the Bookmarks widget crashes it

jsulli:
Yikes, lots of stuff. I'm guessing there are big differences between devices on the full screen setting, because only about half of these are present for me.

But I kind of expected it, I deleted a bunch of the immersion code knowing that'd it'd probably break things. I'll restore most of that.

jsulli:
Yikes, lots of stuff. I'm guessing there are big differences between devices on the full screen setting, because only about half of these are present for me.

But I kind of expected it, I deleted a bunch of the immersion code knowing that'd it'd probably break things. I'll restore most of that.

baka-kaba:
Well for what it's worth I tested it on an emulated 8.0 Pixel 2

baka-kaba:
I should probably point out that this has been completely overhauled! It's still not Kotlin though (and some legacy stuff like the Immersive Mode code is still there) so it can still be redone, there's just a lot less to do (and some of it's been written with Kotlin in mind, like the ``NavigationEvent`` handler)

PR #578, opened 2018-01-16
Remove ICS, JB and zepto

Sereri:
I hope y'all brought your reading glasses cuz it's reading time

Planing to squash the commits so no one sees how bad I am at git-ing.

The javascript might have a few kinks here and there but that's what beta is for, isn't it emot-shepface


Sereri:
Ignore the merge conflict for a second, I can adapt the File before merging

Sereri:
- [ ] timg wrapping the image in a link does not work atm

Sereri:
- [ ] Also Element.closest() is not a thing on 4.4

Sereri:
Welp. After I finished the feature parity I tested this on a KitKat emulator and my 4.4 test phone and it's just not working. Seems like the difference between the Kitkat webview and an Lollipop emulator with play services (does that update the webview nowadays) is bigger than I thought.
4.4 supports almost no part of ES2015 (which coming out in 2014, duh), making me use so many polyfills that it's basically not worth it.
Seems like we're jumping straight to 5.0 after all

PR #583, opened 2018-01-21
Rework main activity

baka-kaba:
Complete overhaul of how the activities work.

A lot of the code in ``ForumsIndexActivity`` has been broken out into other components. There's a viewpager controller that basically manages the forum pages, what's being displayed, and events like "show this page" or "open this thread" (which also involves showing a page). It handles restoration of its own state (including getting references to the existing fragments when it's recreated) so all that complexity is kept in one place. There's still some stuff exposed for the moment (the activity can still get to the fragments) but that might get refined later. It also handles generating a title for the activity, so instead of fragments fighting to change it, the controller can work out what's being shown and produce a title.

Fragments handle their own state (not fully working yet), and are basically supposed to look after themselves - display something, restore when necessary, let the activity know when they've changed so it can react to that event. The activity is mostly a coordinator at this point, passing events around and providing a navigation interface - there are methods like ``showThread``, ``showBookmarks`` etc that anything can trigger

That interface has also been pushed up into ``AwfulActivity``, which basically allows any activity in the app to fire those commands. All the intent handling is done there - there's a new ``NavigationEvent`` type that basically defines things like "show this thread" or "show the forum list", that vastly simplifies the ``onNewIntent`` code

It should be easier to work with, and fixes a lot of bugs, conflicts between fragments trying to set titles and the like. Oh and you should be able to put the ``NavigationDrawer`` into any ``AwfulActivity`` now and have it Just Work (haven't tried it yet though....)


baka-kaba:
Am I ok merging this in? It's not finished but it mostly is, and I don't want people working from an older version or merging might become a nightmare. I left it as a billion commits in case it's terrifying but I'll rebase it a bit before the merge

Sereri:
Maybe branch off a beta branch before you do that since the version before that seems to be working

baka-kaba:
Well I meant roll it into develop so we're all working off the same code (even if some things are temporarily broken)

This looks like it's working apart from the thread list doesn't update when you open a thread (the read count doesn't change), and the thread html doesn't update. Both do if you ``onResume`` them, and it *looks* like I've basically uncovered more spooky magic that was making them work through side effects

The webview just doesn't refresh - all the code runs, the ``bodyHtml`` gets updated and it pulls it in, if you inspect it in DevTools the DOM has changed to the new thread... the display just doesn't change

baka-kaba:
Okay that wasn't spooky behaviour, that was me not noticing onPageVisible did something important. The webview was just paused and the forum view only updates when it's 'visible' (which means the currently selected page, if it's visible in tablet mode but not the current page, it doesn't update)

This looks like it's all working now, aside from fragments not saving and restoring their exact state (scroll positions and the like). That shouldn't take long to do and it's not a showstopper, only comes up when the app recreates anyway

Sereri:
I meant to branch development into a beta branch before merging that into development.  But if it's actually working now you can just merge it and we'll do a beta branch on the next beta release

baka-kaba:
Ohhh ok, I see what you mean now - thought you meant make *this* a beta branch

But yeah looks good now, I'll squash it

Issue #438, opened 2016-08-15
Attempting to scroll on a [code] block brings up the bookmark panel


natchapman:
When scrolling through a thread, if my finger is in a [code] block, the app acts as though I have swiped right and opens up the bookmark panel. I'm running Awful 3.2.3.2 on CM11, using an Amazon Fire Phone. Please let me know if I can provide any additional information.

Sereri:
Could be "solved" by disabling the code-tag swipe-lock at the bottom of a page. But then you could not scroll those very well. vOv

baka-kaba:
I think I have a better idea of what's going on now. The javascript is calling the ``enableSwipe`` method, so its ``touchEvent`` is ending even while your finger is on the element. I'm not sure, but I'm assuming something like this is happening:
- the ``SwipeRefreshLayout`` is allowing touch events to pass through to the webview, so long as it can still scroll
- this lets the webview consume touch events on things like code blocks, and it can block the view pager from consuming horizontal swipes
- but when the webview hits the top or bottom and you're still dragging, the ``SwipeRefreshLayout`` takes over the touch event and stops passing it through, so it can handle the pull-to-refresh gesture
- the webview stops receiving touch events, so the javascript fires its "touch event ended" callbacks and sets swiping enabled on the viewpager
- the viewpager, which has been ignoring motion events up until now, suddenly gets an ``ACTION_MOVE`` event with a horizontal offset (i.e. from where the touch event started)
- it reads this as a sudden horizontal swipe, and basically drags the pager across by that amount

you can see this kind of behaviour by swiping horizontally by different amounts before you swipe up - the pager pops out by about the same amount. Swipe left instead and you get the scroll limit tint thing flying out.

So I'm assuming we can fix this by somehow getting the view pager to handle an in-progress drag better, by taking the first event it sees and treating that (and its position) as the start of the drag. I don't know much about how ``MotionEvent``s work yet but hopefully it's not too hard to do this kind of thing

baka-kaba:
Thinking about it, the viewpager must be seeing that initial ``ACTION_DOWN`` event, because it receives it earlier than the webview (and the javascript's "disable swipe" callback takes a while to happen anyway). So it's probably aware that a drag is in progress, it just has a blackout until swiping is enabled again
- drag starts at ``x, y``
- gets a few motion events at ``x+15, y+10`` or so
- swiping is disabled
- ............
- swiping enabled, next movement event has the finger at ``x+1000, y+2000``
[...]

PR #591, opened 2018-02-01
Fix crash when swiping code blocks, rename and move ToggleViewPager

### LINKED ISSUE #588: Swiping a code block that's at the bottom of a thread can crash the app
If you scroll a thread right to the bottom, and there's a code block there, swiping across it and also moving your finger down can cause this crash
````
java.lang.IllegalStateException: No fake drag in progress. Call beginFakeDrag first.
    at android.support.v4.view.ViewPager.endFakeDrag(ViewPager.java:2549)
    at com.ferg.awfulapp.widget.ToggleViewPager.setSwipeEnabled(ToggleViewPager.java:75)
````
where the code in question is
````
if (enabled && !swipeEnabled) {
    beginFakeDrag();
    endFakeDrag();
}
````
🤔 
maybe it needs a sleep, or an actual ``fakeDragBy()`` call, or something. Related to #438 but this one is a crash instead of an annoyance

baka-kaba:
I've renamed ToggleViewPager to something more descriptive (SwipeLockViewPager)
and moved it into the ForumsPagerController file, since they're related and it's
just a small wrapper that's only used in the view pager anyway.

Slightly reworked the code so the super implementations are only called IF
swiping is enabled - before they were executed anyway and the swipe toggle only
affected the returned result, which seems wrong

Removed the #beginFakeDrag and #endFakeDrag calls since the latter was crashing
the app sometimes, because the fake drag was not in progress

Fixes #588


baka-kaba:
What exactly does the ``beginFakeDrag -> endfakeDrag`` code do again? It seems to work the same without it

We still have the issue in #438 where the viewpager jumps when you swipe up on a code block but the page can't scroll anymore (i.e. the thread's at the bottom of the page). I've put a bit more detail in there about what I think's going on

Sereri:
When the webview disables the swiping it takes a few ms to detect the touch. In this time the viewpager moves just a little bit. When the touch then ends the viewpager aborts the swipe and the page stays where it is at like 90-95% visibility. The fakedrag is there to simulate another swipe which causes the viewpager to reset the page to 100%.

baka-kaba:
Oh ok yeah, I remember that. It doesn't seem to be happening now, possibly because I reworked the disabled swipe logic

I think I've sorted out the jumping issue too, so I'll merge this in

PR #604, opened 2018-04-30
fix postimg.org images

gibb3h:
moved from JS to java image processing by request


Sereri:
Small request, could you add a space after the if and one before the opening curly bracket?

PR #608, opened 2018-05-23
New navigation

baka-kaba:
Ok bear with me here... This is a reworking of the way navigation happens in the app.

Right now we have chains of boilerplate method calls for each action that can occur in various parts of the app (e.g. "showThread" in a fragment that calls "showThread" in the activity that calls "showThread" in the ViewPager...), or we directly create Intents that the resulting Activity has to reinterpret (and then start the chain of method calls again). It's brittle, requires a lot of boilerplate and repeated code, and makes adding new functionality complicated since you have to add a new set of this handler code throughout the app

So instead I've created a message-passing system, where we define a set of NavigationEvents and a basic routing system that allows them to propagate to the component that handles them. Each component (fragments, activities etc) just needs to implement a couple of methods in a NavigationHandler interface - one that handles events it knows about (including passing them to a subcomponent), and one that passes the rest up the hierarchy. An activity that receives an event it can't handle just calls #getIntent on that event - it returns a (completely configured) intent ready to start the activity that handles it. That activity can call a parsing function on its intent, and get a NavigationEvent instance back, ready for routing to the appropriate component. All the complexity is abstracted away - components just need to opt in to handling certain event types, and pass on the rest. NavigationEvents just need to define the activity they want to navigate to, and how to turn themselves and any data into an Intent (and back again)

The idea is that it doesn't matter where an event originates, the routing will deliver it to the appropriate activity and the component that handles it. So you can trigger all kinds of navigation events in the app without having to manually propagate all those method calls - e.g. you can drop the NavigationDrawer in any other activity, so long as the default routing is implemented (which is simple, and already in AwfulActivity/Fragment anyway) its menu items will Just Work

Creating new events should be as simple as defining the activity that handles it, and coding that handling in there - so for example you could add a "show embedding settings" event that targets the Settings activity, and put code in that activity that handles that event by opening the Embedding page, maybe even highlighting an option. Then you can create a popup in the main app that says "hey you can embed things like tweets, want to go to the settings?" and take the user there just by calling Activity.navigate() with your new event

It also means you can define very specific events with associated data - like "show forum" requires a forum ID, but "show bookmarks" doesn't require an ID at all, that's an internal detail for the handler to worry about. This can make handler logic a lot cleaner - right now the Reply/Quote/Edit code does a lot of intent wrangling to work out if a particular parameter is present, what this means, whether something should be treated as a particular type and what data should be present for each. Having actual Event types means you get the benefits of the type system, you know what properties exist etc. For example, a NavigationEvent.Quote could have a list of post IDs to quote (making single- and multi-quote effectively the same thing) whereas NavigationEvent.Reply would only have a thread ID property

Lastly it keeps the technical details all in one place, including serialising and deserialising data as Intents. So Activities don't need to concern themselves with any of this, they can just deal with the event objects. It also means things like Intent configuration can be specified in one place, which is a lot safer, since they involve specific flags that need to be consistent across the app.

The main downside with this system that I can see is that routing can't be validated at compile time like with explicit method calls - you need to make sure that the handling Activity _does_ handle that event, and routes it to something that consumes it. If you don't then the event could be passed around in a loop until you get an overflow. It's a similar limitation to fragments expecting to be in a certain type of activity - you can only tell if there's a problem when the code runs. So it's not ideal, but I don't think it's a huge issue


baka-kaba:
I've been messing with this for a while - it's not perfect and I'd like to refine some things a bit more if possible, but it is working and basically a finished replacement for the current functionality, and I don't want to sit on it since it affects a lot of classes that people might be working on. But it's a pretty big change in how the app is structured, so I didn't want to just merge it without checking it's ok (and yeah it needs squashing a bit)

But yeah, the idea is to make it easier to define actions in the app, that just work no matter where you invoke them, so you don't need to worry about making the navigation happen. It's _kinda_ like an event bus, but explictly routing to a specific consumer, and also handling the whole propagate-through-intents issue when the consumer isn't in the running activity. I hope it makes sense, anyway - it should make things a lot neater and make it easier to do cool stuff, like... interactive app tours for new users! Yes it is possible

Also I changed the way the bookmarks widget works (no hardcoded ID) so you need to recreate it for that to work. We could leave in a hack so the old one is handled too, but I feel like it would be better to  just cut that out of there? Pop up a message saying they need to redo it

Sereri:
Hadn't had time and/or energy to look at this yet but will read through it tomorrow. Jetlag's a bitch.

baka-kaba:
Hey it's cool, no rush! It's a design change so probably better to look at it when you're in the mood

PR #610, opened 2018-06-10
Converted ForumsIndexActivity to kotlin, simplified immersion mode

jsulli:
No functionality changes, just code simplification and reduction

I got the immersion mode implementation down to about 7 lines, I'm not sure why it was so over-complicated - complications with previous version of android, perhaps? I tested it extensively and could find no issues with this implementation. I based it on the sample at https://github.com/googlesamples/android-ImmersiveMode so it should be kosher.

I had to make a minor tweak to Intent.parse in NavigationEvent. As far as I can tell there is no way to call this from kotlin the same way there was from Java, so I made it a normal method that takes an intent as a parameter, instead of an extension method. If there is a way, please let me know!

Removed PagerCallbacks. As far as I can tell, PagerCallbacks had no practical use. It only called setProgress (a deprecated method), which didn't show anything as there is already a custom progress bar implementation in the toolbar. I did pretty extensive testing but didn't find any bugs, so it appears to be a safe removal.

Also updated a handful of dependencies. 

annnddd, as always i mix up which account I'm commiting from.


Awful.apk/src/main/java/com/ferg/awfulapp/ForumsPagerController.kt
```diff
@@ -135,16 +134,6 @@ class ForumsPagerController(
     }
 
 
-    /**
```
baka-kaba: Can you revert the deletions in this file? Even if an activity weren't using these callbacks right now, this is all state management wiring that only exposes actual page changes, deferring them until the page fragment is actually created if necessary. The idea behind this controller class is to keep all the internal viewpager wrangling hidden and managed safely

Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.kt
```diff
@@ -0,0 +1,276 @@
...
+
+import timber.log.Timber
+
+
+class ForumsIndexActivity :
+        AwfulActivity(),
+        PmManager.Listener,
+        AnnouncementsManager.AnnouncementListener {
+
+    private lateinit var forumsPager: ForumsPagerController
```
baka-kaba: lateinit doesn't work for the pager controller and the nav drawer, causes crashes when the app recreates (turn on Don't keep activities in dev options)

Basically the various fragments can get recreated before the activity hits onCreate, which is where these things get initialised. When they call the activity to update it about their state, you need to ensure the nav drawer and pager are ready - there's a race condition when you're restoring the app, and often they're not ready before the fragments

The original version does null checks on those two components, and basically ignores the update call if they're not ready. The forums pager will eventually initialise, it'll recover those fragments and add them to its viewpager, and then it'll inform the activity that the current page has updated through PagerCallbacks (another reason not to delete it!)

So yeah, unfortunately we'll have to keep these nullable and initialise them as null, and do null checks where they existed in the original java version. Most of the other calls will happen after onCreate though, so we can !! those

Awful.apk/src/main/java/com/ferg/awfulapp/NavigationEvent.kt
```diff
@@ -138,38 +138,41 @@ sealed class NavigationEvent(private val extraTypeId: String) {
...
-                )
-                TYPE_THREAD -> Thread(
-                        id = getIntExtra(Constants.THREAD_ID)!!,
-                        page = getIntExtra(Constants.THREAD_PAGE),
-                        postJump = getStringExtra(Constants.THREAD_FRAGMENT)
-                )
-                else -> {
-                    Timber.w("Couldn't parse Intent as NavigationEvent - event key: ${getStringExtra(EVENT_EXTRA_KEY)}")
-                    MainActivity
+        fun parse(intent: Intent): NavigationEvent {
```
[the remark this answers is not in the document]
jsulli: Ahh, I knew it would end up being something simple like that, thanks!

Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.kt
jsulli: Thanks for the note about the dev option. I always forget about that and I definitely need to be testing with it on.

On your recommendation, I mostly agree although now that I know there is a race condition there I'm concerned about other edge cases that could show up - lateinits are really prickly if they aren't safely initialized first. That said, you're way more familiar with the app architecture than me so I'll go with that, with the expectation that it could change in the future. I'm making an `activityInitialized`  field in `AwfulActivity` - I could put it in `ForumsIndexActivity` but it seems like it'll be useful for other potential race conditions like this so it might as well go in the base object, I can move it if you don't agree.

jsulli:
Still clocking in at just under 300 lines after the fixes 🎉 

One last note - so I did revert all the `PagerCallback` changes, but from my testing I still can't see a use for any of the calls in `ForumsIndexActivity::onPageChanged`. Commenting out all the code in it seems to produce no bugs for me, though I'm well aware there could be an edge case I'm missing. The titles all change correctly, and the progress bar shows.

Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
```diff
@@ -40,6 +40,7 @@ abstract class AwfulActivity : AppCompatActivity(), AwfulPreferences.AwfulPrefer
     private var customActivityTitle: TextView? = null
     // TODO: this is a var but honestly why does any activity need to replace it with their own copy? It's a singleton - if they're doing it for the callbacks, just use the register method
     var mPrefs: AwfulPreferences = AwfulPreferences.getInstance()
+    var activityInitialized = false
```
baka-kaba: Can you move this into ``ForumsIndexActivity``? It might be good to have something like this here at some point, but this flag's being used to guard the lateinit stuff, and it's being set before those things are actually initialised. It _works_ fine as things are, I'm just worried about it introducing subtle bugs down the line

I'd put it after the navigation drawer is initialised in ``onCreate``, but before the ``updateNavDrawer`` call that relies on it. And can you stick a comment above that just says the components are all initialised at this point? Maybe things need reordering in there but this'll do for now

baka-kaba:
Also you don't have to do this if you don't want, but for a lark we could do

````
        return when (event) {
            is NavigationEvent.MainActivity ->
                true
            is NavigationEvent.ForumIndex ->
                true.also { forumsPager.currentPagerItem = Pages.ForumIndex }
            is NavigationEvent.Bookmarks, is NavigationEvent.Forum, is NavigationEvent.Thread, is NavigationEvent.Url ->
                true.also { forumsPager.navigate(event) }
            is NavigationEvent.ReAuthenticate ->
                true.also { Authentication.reAuthenticate(this) }
            else ->
                false
        }
````
for the ``handleNavigation`` method. I think it looks pretty neat! But it's not a big deal

jsulli:
Ohhh, never seen that `true.also {}` trick, that's neat, definitely gonna do that.

Also, now I do see what you mean about `updateTitle()` in `onPageChanged()` (took a couple repeats for it to actually fail to update), however it still seems like there is no need for `setProgress` in there? Sorry to keep harping on it :v

jsulli:
Also, be careful when trying to use `NavigationEvent.Thread` directly. It's easy for Java/Kotlin to get confused and think you mean `java.lang.thread`, heh

baka-kaba:
Eh I'd leave ``setProgress`` - it shows the loading bar for the fragment the user's paged to. I know it's deprecated and it doesn't exactly show anything useful (except "still waiting on some request"), but it's still our visual loading indicator and we don't have anything to replace it with yet

And yeah there are other classes called Thread floating around too 👀 we can change it or rename things if it gets awkward

jsulli:
Thanks for the detailed reviews as always, it's really helpful.

Once this is merged in I think I'll move on to the next beast, `ForumDisplayFragment`, unless you have a specific target you want cleaned up.

baka-kaba:
Hey no problem, thanks for putting up with the demands! 🐢

``ForumDisplayFragment`` though... I've redone ``AwfulThread`` and ``AwfulPost`` as more sensible (Kotlin) classes, and there are Thread and Post repos to abstract all the ``ContentValues`` wrangling and requests. So that fragment's code will end up pretty different. It's up to you, but I think it'll probably be better to leave it for now - I'll try and push that feature branch soon though, it's basically done I think? (``ThreadDisplayFragment`` has been completely rewritten too)

You could have a crack at ``PostReplyFragment`` if you like! It's pretty big (and it'll need testing since there's a bunch of important functionality) but it can definitely be cleaned up. I want to rework that with ``Post`` and ``Quote`` and ``Edit`` ``NavigationEvent``s instead of ``Intent``s at some point too. Or you could do any of the smaller classes, or have a go at things in the Issues list? Someone needs to sort out the backup system (it backs up the cache database and if that's too huge it won't save your settings at all)

Issue #616, opened 2018-06-15
Thread display is reverting its scroll position in #onResume


baka-kaba:
If you load a page that jumps to a position (a specific post or the first unread one on the page), scroll it a bit, then do anything that pauses the fragment (open a link and return, switch task, pull up the share menu etc) then it'll reset the scroll to that jump position when the fragment resumes.

They've [rewritten Loaders in the support library](https://developer.android.com/topic/libraries/support-library/revisions#27-1-0) to act more like ``LiveData``, which includes reloading the data when an Activity/Fragment hits ``#onStart``. Seems to be [intended behaviour](https://issuetracker.google.com/issues/74278849) so we'll have to deal with it

The problem is the app is written with the expectation that a loader only runs when there's an explicit request for a page load/reload. It jumps to the target post or scrolls to the first unread one because that's standard behaviour for loading a page. But triggering when the fragment regains focus is more like refreshing the underlying data, it shouldnt cause any navigation (especially since the underlying data probably hasn't changed at all). And it's hard to untangle this from the way the app currently works

I'm rewriting the whole fragment and removing loaders anyway, but for now we just need a fix that stops it from being annoying. I think the best idea is to try and store and restore the scroll value in ``onPause``, so it's always acting like the fragment is being recreated (with its last state)

baka-kaba:
Wow this is just the worst. The thin line between order and chaos
It might be quicker to just finish my rewrite

PR #625, opened 2018-06-30
Revert support library to 27.0.2 to avoid a breaking change

baka-kaba:
See Issue #616
They changed the behaviour of Loaders, including forcing them to reload
when a fragment or activity hits #onStart
Changes:
https://developer.android.com/topic/libraries/support-library/revisions#27-1-0
Intended behavior / WONTFIX:
https://issuetracker.google.com/issues/74278849

This makes the thread page reload every time the user switches away from
the app, or visits a link and returns, turns the screen off... page
loading isn't designed to handle this, and if it was initially loaded
with a post to jump to (either a post ID fragment or "jump to last read")
it'll jump back there again, every time. It makes catching up with a
thread a really bad experience

A real fix would require an overhaul - I can't see any reliable way to
block these specific reloads and not others (e.g. recreating a fragment)
without a lot of additional complexity. Vetoing the scroll behaviour
would require keeping track of when a post jump has been "consumed" so
it doesn't happen more than once, but it also needs to be able to identify
a "new" load - if the user hits a "jump to post" link that scrolls
somewhere in the same page, and then scrolls down, you don't want it to
scroll back if they navigate away and back to the app. But if they scroll
to the jump link again and hit it, the scroll needs to happen this time

Scrolling in the WebView needs fixing in general - right now there's no
way to get and restore the current scroll position, so we can't really
recreate fragment state

### DIFF
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -118,8 +118,9 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
 
-    implementation 'com.android.support:appcompat-v7:27.1.1'
-    implementation 'com.android.support:design:27.1.1'
+    // held back to fix Issue #616 - breaking Loader change introduced in 27.1.0
+    implementation 'com.android.support:appcompat-v7:27.0.2'
+    implementation 'com.android.support:design:27.0.2'
 
     // used to fix SSL issues on older devices
     implementation 'com.google.android.gms:play-services-auth:15.0.1'


baka-kaba:
Is this going to cause any problems? The behaviour change really hits this house of cards and it's just easier to revert for now, until all the underlying issues are fixed. And this way we're not holding up another release at least

Sereri:
Seems fine to me, we should try to work around the feature change though so we don't have to stay on this version forever.

baka-kaba:
Looks like I needed to revert the design lib too...

Yeah this is a temporary thing, just wanted to make sure it wouldn't break anything anyone was working on. I added an issue about some changes we need to make

PR #628, opened 2018-08-06
Thread view overhaul

Sereri:
Technology changed in the last 5 years, who knew. Cleaned up a good chunk, put all (or most) duplicate CSS into general.


baka-kaba:
Looks good! I'll dogfood it for a bit and see if I see anything weird

While you're at it, could you make the touch target on the post menu ... glyph bigger? It feels hard to hit sometimes, like I need to poke it multiple times - I'm guessing the detection box just wraps the glyph so it's tiny?

Also out of interest, are there any cheap out-of-the-box animation transitions we can throw on the expanding user details bit? It would be nice if it slid open and appearing text faded in, the avatars growing to their larger size, that kind of thing. No biggie but if there are some easy drop-in HTML5 functions for that kind of eye candy, that would be cool

Sereri:
emot-allears

baka-kaba:
What the hell! I'm sure that wasn't happening yesterday...

That looks rad

baka-kaba:
There's a thick line above every post, which kinda makes it look like the start of a quote. Is that meant to be there, or is it meant to replace the full-width line under the post details section?

Sereri:
Can you screenshot that?

baka-kaba:
image

It's this rule in ``general.css`` that's doing it
``border-top: 2px solid #eae9ea;``

baka-kaba:
You removed the new line (that's in the other themes) and left the full width line in OLED, I'm guessing you didn't mean to do that!

Also the text isn't visible in post previews on the OLED themes I tried (OLED, YOSPOS) and the line pops up in some of them (those plus Dark), I haven't tested everything though

Menu button seems a lot better!

Sereri:
> Having used it a bit, there's a couple things I noticed. Avatars look ok flush against the left side of the screen, but when someone doesn't have one, their username is right up against the edge and it looks a little weird to me. Maybe add the padding to the left of the name, so whether there's an av or not it's always offset?
I forgot to add padding there, will fix

> The other thing is a bit weird - sometimes I get what looks like a single line of an image from an earlier post, at the bottom of a later post (say 3 or 4 posts below). It can be a picture someone posted, or an image in an embedded tweet. Sometimes it stays if you scroll it off and on the screen again, other times it changes or disappears. 
I think that might be a problem with hardware-acceleration. I'll look into that. Can you try to get a screenshot next time?
>The bottom of embedded tweets sometimes cuts off too - the border, so you get the rounded corners and then its just cuts off. That must be more than one line, but it might be related?
That's because they get loaded later and don't calculate the height correctly. Has been a problem since the beginning.

baka-kaba:
I'll give it a go with the screenshot, it only happens occasionally. Or maybe it happens all the time, it's just that most posts have blank areas so you can't see the bit that's appearing? Is there like a one-pixel gap in the layout or something? Like now the posts fill the screen, is there any background or is it just the webview void?

Also (also) I reckon lists could do with being less indented. Maybe like a standard indent for all block content, like 16 or 24dp or something?

baka-kaba:
Ok here's some yospics

image

Little slice right at the bottom of the last post on the page - it's attached to it, so it follows when you scroll up and down. Scroll it off the page and it might disappear/change. Looks like it came from this post a few up:

image

Not just at the end of the page though, so maybe there's a gap after every post with the void poking through?

image

Sereri:
I rolled back applying a default / dark fix to everything so the bug should not be happening anymore. I also decreased the quote indentation from 20 to 15. The twitter thing is a bug that happens from time to time with it. It tends to fix itself most of the time (i.e. twitter fixes their shit).

PR #629, opened 2018-08-11
Enable fullscreen videos in the WebView

baka-kaba:
Creates a fullscreen dialog and sets the view passed by the webview as
the dialog's contents. This sets the system UI flags to the standard
fullscreen/immersive-sticky setup, and resets them when the activity
regains focus.

There are a couple of issues - one is that the fullscreen dialog is
sometimes blank, but the content is there and you can still interact
with it - you just can't see it. Swiping on the screen makes it appear.
The main issue is that the navigation and status bars remain hidden
after the UI flags are reset, until the userdoes something to bring them
back (poking those areas, flipping the viewpager, rotating the screen
etc).

I've added a handler and a runnable to post to it that forces the
"show UI" flags on a delay, but the delay doesn't seem to make any
difference to when the flag is applied. It doesn't work without setting
the flag, but it can take effect and the UI can pop back in at any time,
before the runnable itself is executed. It Is A Mystery!

So sometimes it behaves and the layout has space ready for the UI to
neatly slide back in, and sometimes the UI appears over the layout and
it redraws and looks janky. I've tried a lot and it seems like you can
either have this inconsistent behaviour, or leave the UI hidden until
the user does something to bring it back.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.kt
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ForumsIndexActivity.kt
@@ -36,11 +36,13 @@ import android.content.Intent
 import android.content.res.Configuration
 import android.net.Uri
 import android.os.Bundle
+import android.os.Handler
 import android.support.design.widget.Snackbar
 import android.support.v7.widget.Toolbar
 import android.view.KeyEvent
-import android.view.View
-
+import android.view.View.*
+import com.ferg.awfulapp.NavigationEvent.*
+import com.ferg.awfulapp.NavigationEvent.Companion.parse
 import com.ferg.awfulapp.announcements.AnnouncementsManager
 import com.ferg.awfulapp.constants.Constants
 import com.ferg.awfulapp.dialog.Changelog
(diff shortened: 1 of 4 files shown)


baka-kaba:
All right, I can't make this work cleanly - if you just set the flags the UI stays hidden, if you post a runnable setting the Show UI flag it works but completely separate from the runnable being executed, so it sometimes pops up over the layout. If you do a runnable without the flag setting code it doesn't work. If you put that flag-setting runnable in a different execution branch it still works. I have nooooo idea what's happening and I'm pretty bored of it now!

So yeah our options are
* returning from fullscreen leaves the UI missing, the user needs to make it reappear
* the UI reappears automatically but it sometimes glitches for a moment (this is what's in the PR)

Sereri:
Works for me vOv, I guess we can go back to the other variant if people in the beta have troubles. Not sure if this is a Moto thing but rotating the screen on a playing video automatically makes it fullscreen, rotating back goes back. This worked on the old build as well and also restored the nav bar. I guess due to the rerendering on rotate

baka-kaba:
Nah I don't get that unfortunately so maybe it is a Moto thing, or a newer Android thing - I have rotation locked, but if people are gonna rotate to watch a fullscreen video (pretty much the main reason to go fullscreen anyway) then yeah it'll bring the UI back just from rotating back again, the activity layout gets a kick

I'll merge it then!

PR #632, opened 2018-08-18
Leper's colony

baka-kaba:
Activity that lets you browse the LC and view users' rap sheets.
Basically shoehorned into the thread view, so all the funny web content works, and the metadata and navigation options are in the header.
Uses some thread icons for ban reasons right now, it would be good to have some specific ones for all the possible ban types, but the name is there when you expand the header anyway.

It's still a work in progress, but I figured I'd make it a PR so y'all can try it and see if it needs anything. There are some TODOs in the activity, mostly navigation stuff


baka-kaba:
I dropped the phone guys for some plain requested/approved labels like this:
image
that's with the system text size set to max, so you can see how it wraps for longer names (since the icons were smaller), and how the extra lines wrap below the icon/avatar area

Is this ok for now? We could have the "requested by" etc parts on their own lines, but I feel like those would need some CSS changes to make them a little smaller or something, so there's not a big stack of text. The wrapping issue would need to be changed in the main layout CSS to put the avatar in its own column

Or we could move the requested/approved/verdict bit below the icon, and maybe right-justify it (so when you have the same requester/approver the names line up - "requested" and "approved" are different lengths, so when they're on the same line as the name they end up offset a bit)

baka-kaba:
Only thing with moving the verdict up first is it kinda reads like "BAN requested by: Some Overworked Mod" and it might imply that the mod is responsible for the punishment. I mean I don't know how it works and maybe they do ask for bans, but I wouldn't want to give the wrong impression and get people blamed, y'know? I should ask someone

I can put the extra colons in if you want, I had them at first but I think this way reads a little more natural and friendly. ``Verdict:`` has one because it is _very serious_

Sereri:
I hadn't thought about that, good point. If the A/R move down below the avatar there would maybe be a disconnect?

Sereri:
I'm liking all the changes, the autoban icon looks great. Menu position sound right. I propose the biohazard icon again or maybe something similar

baka-kaba:
Almost done with this, had to make it a fragment and crowbar it into the AwfulFragment system so we get the PTR and so on. There's a bug I need to squash too, when the link to the nuaghty post isn't actually a link

baka-kaba:
All right, this is done - it's just a single fragment that does double-duty as the Leper's Colony and a displayed rap sheet. It might be better as two fragments so the activity can do a master/detail view, keep scroll positions, show the rap sheet as a popup etc but I'm pretty done with this for now

Anything important before I merge it?

Sereri:
Nope, if theres anything we'll fix it in post

Issue #627, opened 2018-07-12
Convert link-wrapped images correctly when images are turned off


baka-kaba:
I think this is a separate issue from #618 really - currently when the user has images off, any ``<img>`` tags are converted into ``<a>``, with ``href`` set to the URL of the original image. This lets the user click the link to go and view the image (and the app will actually allow it to be loaded and displayed inline).

But if the original image was already wrapped in an ``<a>`` tag, the resulting link points to that URL, not the URL of the image. It looks like the original link is retained, and the image URL becomes text in a ``<p>`` tag wrapped by that anchor tag.

So we need a way to handle this parsing situation, turning the image into link while also retaining the original link. It might be best to create a pair of links, maybe with an icon to show what it relates to:
🖼 [full image url]()

🔗 [full link url]()
(with enough space that they can press the right one easily)

Also need to decide if clicking the image link and doing "display inline" should also rewrite the other link to wrap the image, as it originally was. A little more complicated, but maybe we could use some custom CSS classes to tag converted images and their links, so they're associated and it's easy to reverse the process

PR #633, opened 2018-08-25
Fix for #627 - Link wrapped images

mensab:


Awful.apk/src/main/java/com/ferg/awfulapp/thread/AwfulPost.java
```diff
@@ -525,7 +525,20 @@ public static void processPostImage(Element img, boolean isOldImage, AwfulPrefer
                 // switch out for a link with the url
                 img.replaceWith(new Element(Tag.valueOf("a"), "").attr("href", originalUrl).text(originalUrl));
             } else {
-                img.replaceWith(new Element(Tag.valueOf("p"), "").text(originalUrl));
+                if (alreadyLinked) {
+                    Element parent = img.parent();
+
+                    parent.appendText(img.parent().attr("href"));
```
Sereri: `parent.appendText(parent.attr("href"));`

baka-kaba:
We need some CSS styling for this - from a little messing around I got this:

image

with this in ``general.css``
````
.converted-to-link {
    display: block;
    border: 2px #00c0f4 solid;
    padding: 8px;
}

.converted-to-link::before {
	content: "Image converted to link";
	font-style: italic;
	font-size: 0.8em;
}

.converted-to-link a::before {
	padding-right: 2px
}

.converted-to-link a {
	display: block
}

.a-link {
    padding-bottom: 1.5em;
}

.a-link::before {
	content: "🔗";
}

.img-link::before {
	content: "🖼";
}
````

The padding needs adjusting and honestly, I think material icons for the link and image would look way better (and my CSS is bad) but how's that looking? Also for consistency I think it would be a good idea to wrap the converted image too, like this?

image

And I think it might be better to have the image url first, _then_ the ``a`` link - I feel like the image is the "main" element, yknow?

mensab:
I've updated it to put the image url before the a link.

My css skills are non-existent.

baka-kaba:
I just poke at the CSS in the ~wallet inspector~ device inspector in Chrome. Sereri's reworking that stuff at the moment though!

baka-kaba:
Looks good and seems to work!

I tweaked the CSS since the elements are switched around - this should work for plain images that are converted to links too (if we add the classes to those as well)

````
.converted-to-link {
    display: block;
    border: 2px #00c0f4 solid;
    padding: 8px;
}

.converted-to-link::before {
	content: "Image converted to link";
	font-style: italic;
	font-size: 0.8em;
}

.converted-to-link a::before {
	padding-right: 2px
}

.converted-to-link a {
	display: block;
	padding-top: 8px;
	padding-bottom: 8px;
}

.a-link::before {
	content: "🔗";
}

.img-link::before {
	content: "🖼";
}
````

PR #635, opened 2018-09-25
Update target API to 28, and libraries and plugins

baka-kaba:
- API 28 set as target/compile version
- Support libs updated to version 28
- Gradle plugin to 3.2.0
- Kotlin plugin to 12.7.1
- Play Services Auth to 16.0.0
- GSON to 2.8.2

also fixed some nullability errors that were introduced

### DIFF
--- a/Awful.apk/build.gradle
+++ b/Awful.apk/build.gradle
@@ -124,14 +124,14 @@ dependencies {
     implementation 'com.android.support:support-v4:28.0.0'
 
     // used to fix SSL issues on older devices
-    implementation 'com.google.android.gms:play-services-auth:16.0.0'
+    implementation 'com.google.android.gms:play-services-auth:16.0.1'
     //implementation 'com.mcxiaoke.volley:library:1.0.19'
     implementation 'com.github.samkirton:android-volley:9aba4f5f86'
-    implementation 'com.google.code.gson:gson:2.8.2'
+    implementation 'com.google.code.gson:gson:2.8.5'
 
-    implementation 'org.jsoup:jsoup:1.11.2'
-    implementation 'com.jakewharton.threetenabp:threetenabp:1.0.4'
-    implementation 'com.samskivert:jmustache:1.13'
+    implementation 'org.jsoup:jsoup:1.11.3'
+    implementation 'com.jakewharton.threetenabp:threetenabp:1.1.0'
+    implementation 'com.samskivert:jmustache:1.14'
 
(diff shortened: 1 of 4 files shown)


baka-kaba:
Gonna test this a little more, but it's here in case anyone was going to do it themselves

baka-kaba:
All right, this seems fine apart from one weird issue. @Sereri can you work your Javascript magic with this?

Click on a thread that takes you to, say, the last post (one with 0 unread). If you try and scroll up, the page keeps resetting the position, dragging back down, so you're fighting against it. As far as I can tell this is down to the scroll handling in ``thread.js``, which sets up a chain of 100 or 200 scroll adjustment callbacks 200ms apart, so this re-scrolling behaviour takes 20-40 seconds to finish

````
function scrollLastRead() {
    ...
		window.topScrollCount = 100;
		window.scrollTo(0, window.topScrollPos);
		window.topScrollID = window.setTimeout(scrollUpdate, 500);
    ...
}

function scrollUpdate() {
    ...
			window.topScrollPos = newPosition;
			window.topScrollCount--;
			window.topScrollID = window.setTimeout(scrollUpdate, 200);
    ...
}
````

The weird thing is this still happens in earlier versions (you can inspect the webview in Chrome and see ``topScrollCount`` counting down for the whole duration), it just doesn't seem to affect the webview. Scrolling feels fine. But after updating, it looks like something has changed? I'd have thought anything like that would be baked into the webview itself but guess not

Sereri:
I think I got this fixed, will push an update after merging this.

Is there a reason you didn't update
- jsoup 1.11.2 -> 1.11.3
- play-services-auth:16.0.0 -> 16.0.1
- timber:4.6.0 -> timber:4.7.0

Sereri:
Actually, here are the changes
[scrollFix.zip](https://github.com/Awful/Awful.apk/files/2442853/scrollFix.zip)

baka-kaba:
Oh nah, I didn't realise you needed to run an inspection for updates - usually the linter just highlights everything with a newer version? The only one I intentionally left back was the HTTP library since I'm not sure if it'll break anything

I couldn't apply your patch though, I tried doing it by hand and I think you're using a different version of ``thread.js`` - there's a function in your patch diff called "waitForRedraw" or something like that, I don't have that anywhere. I updated the libraries anyway and they seem ok - if you like I can just squash and merge this and then you can push the fix as another commit?

baka-kaba:
My bad, didn't realise you'd pushed to develop. Thanks notifications!

I'll rebase and squish this

Issue #615, opened 2018-06-15
Activate twitter embed by default


Sereri:
If I have to read "why doesn't this app have twitter embeds" one more time I might have to murder someone.

PR #638, opened 2018-09-26
Changing preference defaults.

MaximoB:
I changed the preference defaults to what I personally prefer. This includes: enabling twitter embeds (#615), enabling YouTube embeds, enabling WebM embeds, disabling WebMs autoplaying, and also enabling GIFs. (also lol vines)


reiser-mobizcorp:
Please revert enabling gifs. They are terrible and also terrible for battery life. They should only by the user at their own risk

Sereri:
Please revert enabling gifs. They are terrible and also terrible for battery life. They should only by the user at their own risk

PR #639, opened 2018-10-21
Make launcher icon selectable

Sereri:


Sereri:
Is this better now? (Also lint-resolved something unrelated by accident but left it in)

Awful.apk/src/main/java/com/ferg/awfulapp/NavigationEvent.kt
```diff
@@ -28,7 +29,7 @@ sealed class NavigationEvent(private val extraTypeId: String) {
      * Defaults to opening the main activity, override this for events handled by other activities.
      */
     protected open fun activityIntent(context: Context): Intent =
-            context.intentFor(ForumsIndexActivity::class.java).addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+            context.intentFor("com.ferg.awfulapp.ForumsIndexActivity."+AwfulPreferences.getInstance().launcherIcon).addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
```
baka-kaba: I'd revert all the changes in NavigationEvent - it works without them, and it feels like involving the launcher icon and aliases is adding complexity that could cause weird issues later. Unless there's a reason to keep these changes I'd avoid it, honestly

baka-kaba:
All looks good apart from that change! Also I think the reverted code is what lets you keep a bookmarks widget with a different icon, which people might be into - it always seemed to disappear along with the app icon on your version, not sure why

It would be cool to have a listview in the icon chooser dialog (just items with a graphic and title), but this is good enough to put out there, right?

baka-kaba:
Also I noticed the icon in the switcher is still ``ic_launcher.png`` - I'm guessing there's nothing we can do about that huh

Sereri:
I actually noticed another problem, the app has no title in the task switcher :psyduck:

PR #644, opened 2018-11-13
Remove hard failures when probed

baka-kaba:
I've removed the checks that cause response handlers to fail if the
user is on probation, because it stops pages from loading even when the
user should be able to use them.

For some reason, being on probation produces an AwfulError, even though
it's basically a state and not indicative of a site error etc. This is
recognised through an #isCritical check on AwfulError, which basically
whitelists PROBATION errors as being fine, and not something that should
halt response handling.

But then all the response handlers explicitly fail if #isCritical returns
true OR it's a probation. And I can't see any reason to do that, except
where probations actually block the user from using something, e.g.
composing a reply. And even then, it's better to handle it directly
instead of just letting the action fail

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/task/PreviewEditRequest.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/task/PreviewEditRequest.java
@@ -54,6 +54,6 @@ public class PreviewEditRequest extends AwfulRequest<String> {
     @Override
     protected boolean handleError(AwfulError error, Document doc) {
         Timber.e(error);
-        return error.getErrorCode() == AwfulError.ERROR_PROBATION || error.isCritical();//Don't allow probation to pass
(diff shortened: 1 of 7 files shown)


baka-kaba:
@Sereri do you know if there's any reason to not remove these? Everything seems to work fine, I don't get it

Sereri:
It's been quite a while since I've looked at probation code. I think the main idea was to prevent probatees from accessing the reply window in any case as they couldn't post something anyway.
There might've been copy/paste mishaps on my part in the search requests. If not I have no idea what I was thinking. Thank god for all the comments I wrote oh wait

baka-kaba:
I tried searching and that works, the reply window just pops up the error toast, waits a minute, then closes the window and goes back to the thread. And the original code goes back to the dawn of time anyway so you were probably just copying that!

I'll merge it then, seems foine to me 👍

PR #645, opened 2018-11-13
Add more performant AwfulRequest class to speed up page parsing

### LINKED ISSUE #643: Parsing something with a large page selector dropdown is slow as heck
So when a page has a dropdown to choose from all the page numbers, and those page numbers are in the _thousands_, parsing it into a ``Document`` can get very slow. jsoup has to parse each of those ``option`` elements and turn them into a tree node, and it really adds up

On my device the Leper's Colony (with 3680 pages) takes **~4 seconds** to handle the network response - the actual request handling code that works on the parsed ``Document`` is about 120ms of that. A popular thread with over 6000 pages takes **9 seconds** to process a single page. Obviously this is device dependent, and my emulator runs quicker, but it's still multiple seconds of overhead that scales with the number of pages the thread or whatever happens to contain

Right now all ``AwfulRequest``s automatically parse the resulting page with jsoup (this is the slow part), do progress callbacks and error handling, and hand off the resulting ``Document`` to the implementing ``Request``'s result handler. Stripping out the thousands of ``<option>`` tags before handing it over to the parser would be quicker, but some handlers rely on those elements being there (like when searching for the one tagged as "current page").

So we'd need a way to make that opt-in (set a ``stripBadStuff`` flag or whatever when defining a new request type) and still provide things like the current page number to the handler. Maybe define a couple of response types (``Full`` that provides the full ``Document``, ``Fast`` that provides the edited ``Document`` and whatever pieces of data need to be salvaged) - then have the ``Request`` specify which generic type it implements

baka-kaba:
When a request's response is handled, AwfulRequest uses jsoup to parse
the page data into a HTML Document, which involves building a tree of
nodes representing all the elements on the page. But the pages are fairly
complex structurally, and a big problem is the page-picker dropdowns -
they have an <option> element for every page in the thread/forum etc.
When this is a large number (in the thousands) each page parse can take
a very long time, several seconds between receiving the network response
and handling the document.

This PR uses a regex to strip out the page selector elements before the
HTML is parsed, resulting in a ~10x speedup when there are a significant
number of pages.

It basically works by adding a new abstract subclass of AwfulRequest,
called AwfulStrippedRequest. If you build a concrete request class from
this, there's a new handler method called #handleStrippedResponse, which
passes in a Document as usual (but with the page selectors removed), as
well as the current and last page numbers parsed from the page selector
(since this is important information and probably the only reason you'd
want the page selectors present in the Document anyway)

Right now only 3 requests implement it - the forum, thread and Leper's
Colony requests, which are the places you're likely to see large page
counts. Ideally every request would use it, but that means you need to
be aware that you aren't getting the full forum page. But we could work
something out, with e.g. opt-in flags if you *need* the entire structure,
so there's only one AwfulRequest base class

fixes #643

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/constants/Constants.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/constants/Constants.java
@@ -32,6 +32,8 @@ import com.ferg.awfulapp.BuildConfig;
 public class Constants {
     public static final boolean DEBUG = BuildConfig.DEBUG;
 
+    public static final String SITE_HTML_ENCODING = "CP1252";
+
     //public static final String BASE_URL = "http://forums.somethingawful.com";
     public static final String BASE_URL = "https://forums.somethingawful.com";
 
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/task/AwfulRequest.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/task/AwfulRequest.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.net.Uri;
 import android.os.Handler;
 import android.os.Looper;
+import android.support.annotation.NonNull;
 import android.widget.Toast;
 
 import com.android.volley.AuthFailureError;
@@ -20,7 +21,6 @@ import com.android.volley.toolbox.HttpHeaderParser;
 import com.crashlytics.android.Crashlytics;
 import com.ferg.awfulapp.AwfulApplication;
 import com.ferg.awfulapp.R;
-import com.ferg.awfulapp.constants.Constants;
 import com.ferg.awfulapp.network.NetworkUtils;
 import com.ferg.awfulapp.preferences.AwfulPreferences;
 import com.ferg.awfulapp.util.AwfulError;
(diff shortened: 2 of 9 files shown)


baka-kaba:
All right, what do you think about this sort of thing?

It breaks from the idea that the jsoup parsed page is the exact document you can look at in the Chrome inspector or whatever, but it's only removing one element (well a pair of them), and requesting the full document without the efficiency can always be an option. It's just a little thing to be aware of

It makes a noticeable difference for me on big threads etc - it's that delay going from 25% to 50% on the loading bar, after the network response has been received. It makes pages display pretty much instantaneously. There's probably more we could strip out, but I want to keep that to a minimum so it doesn't get too fragile, and the page selectors are the worst thing on there

It's not totally done, but I wanted to see if this is an ok thing before I go cleaning it up and whatever

Sereri:
Seems like a good idea. It's stupid that the forums still generate all that garbage instead of generating it with JS. I wouldn't have thought that it takes that much time to parse though.
Might be interesting to find out at which point it makes sense to strip the garbage and at which points its fast to just eat the pennies. 100 pages? 1000?

Sereri:
Yeah if it's negligible it makes sense to use it as the default for anything with pagination.

baka-kaba:
I'll merge this since it works, I'm cleaning up all the requests so that can be a separate PR

Issue #631, opened 2018-08-14
Pages don't rescroll to the target post after content has loaded


baka-kaba:
I assume that's the intended behaviour? When you load a page at a certain post (link to a post, last-read etc) and content like images and embeds end up moving the layout as they pop in, when they're done the page is meant to re-scroll to where it was meant to be ?

There are (were) ``WebViewClient`` functions using the ``#onPageFinished`` callback to (I assume) reset the scroll position after content finished loading, by calling the method to display the page again. I guess all the content would be cached and it would just behave itself this time?

Anyhow that callback only happens when the template content is loaded into the webview - actual pages are displayed by replacing the inner HTML of the container div, and that doesn't trigger any loading/finished callback that I can tell.

-------------------------------------

This probably won't come up for most people - loading a new page puts you at the top, and all the content expands _below_ so the view position won't change. Loading a page with read and unread posts also puts you at the top, if you have the 'hide read posts' folding active, which is on by default? It mostly happens if you click a link to a post, either in a quote or an actual link

I think the best option would be something like
* create a callback in the JS for when all the content has finished loading (if this is possible)
* use that to trigger a rescroll to where the view should be, instead of reloading the html into the container
* do nothing if the user has already scrolled the screen manually (because it's annoying to have the page drag you back when you've scrolled away from the entry point)

PR #651, opened 2018-11-28
Webview fixes

### LINKED ISSUE #650: Retry if Twitter embed.js fails to load
Currently we try to embed the Twitter embed.js once when we build the webview. If you're... going through a tunnel or whatnot this might time out and all your tweeps are not rendering correctly.

In my mind it would be ideal to retry on every new page load until it loads.

Sereri:
Fixes #650 and #631


Sereri:
Actually I don't think this fixes #631

PR #657, opened 2018-12-02
Remove roundIcons as they overwrite icons where supported.

Sereri:


Sereri:
see https://medium.com/@ianhlake/on-a-device-that-uses-roundicon-that-roundicon-will-always-be-used-youll-note-that-there-are-ef59a2df4c60

Sereri:
Actually, that's stupid, I can probably  just put the same xml as the normal one into the rounded folder and have it work just the same.

PR #673, opened 2019-01-28
Fix cleartext content in API 28+, use stricter mixed content policy

baka-kaba:
HTTP content (rather than HTTPS) used to be the default in Android, but
in API 28 it's become opt-in instead of opt-out. I've explicitly allowed
it in the manifest.

I've also switched the WebView's MixedContentMode from ALWAYS_ALLOW to
COMPATIBILITY_MODE. The latter should still allow cleartext content that
people post, like images, while blocking any dodgier filetypes. It always
felt like a security hole leaving it on "allow anything", ideally this
will let the WebView people manage the security side without being
restrictive

### DIFF
--- a/Awful.apk/src/main/AndroidManifest.xml
+++ b/Awful.apk/src/main/AndroidManifest.xml
@@ -23,7 +23,9 @@
         android:hardwareAccelerated="true"
         android:icon="@mipmap/ic_launcher"
         android:roundIcon="@mipmap/ic_launcher"
-        android:theme="@style/Theme.AwfulTheme.Launcher">
+        android:theme="@style/Theme.AwfulTheme.Launcher"
+        android:usesCleartextTraffic="true"
+        >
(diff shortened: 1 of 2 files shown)


baka-kaba:
Hey @Sereri can you have a quick check this doesn't break anything on your device before I pull it in?

They [changed this in Pie](https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted) but nobody seems to have mentioned a problem yet - but when someone linked [this thread](https://forums.somethingawful.com/showthread.php?threadid=3793286&pagenumber=1) as an example none of the OP's images showed up on my emulator

Also are we cool with making the mixed content mode more restrictive? I've used it a while and haven't seen any issues, and they say the compatibility mode is meant to act like a web browser, so it shouldn't block anything we want to show

Sereri:
works and seems to have no issues that I noticed

PR #679, opened 2019-02-01
Cookies

wardog1uk:
I've moved all the cookie stuff from NetworkUtils to a new CookieController class. Which just makes them both easier to read.

Also did some work refactoring on the cookie code. The only real change is that before the session id wasn't being saved to a cookie, which I assume was wrong, but revert 277b7d7 if not.


Sereri:
Looks good? @baka-kaba ?

Sereri:
Sorry @wardog1uk 
for some reason I was sure this was merged

PR #680, opened 2019-03-11
Fixed custom font not being displayed in the threads

wardog1uk:


wardog1uk:
Ok this makes posts and messages use the selected font instead of just the default one. We could probably add some other public domain fonts too.

Sereri:
Sorry for asking but what is the problem with the webprovider exactly?

wardog1uk:
It's not working. From what I can see first the debug app tries to access _com.ferg.awfulapp.webprovider_ when the correct url is _com.ferg.awfulapp.debug.webprovider_

But if you change it to that you get:
`2019-03-12 11:03:34.882 29582-29582/com.ferg.awfulapp.debug D/Web Console: Access to Font at 'content://com.ferg.awfulapp.debug.webprovider/fonts/comic.ttf.mp3' from origin 'https://forums.somethingawful.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://forums.somethingawful.com' is therefore not allowed access. -- From line 0 of https://forums.somethingawful.com/
`

I've tried a few things but I've never really used contentproviders much so I might be missing something. Changing to accessing the asset files directly seems to work and you could remove AwfulWebProvider altogether.

baka-kaba:
Wow, now that is a vintage class right there

Honestly I feel like maybe we could revisit the whole font thing now that Android has some functionality for loading and storing fonts on the device, the way we do it is pretty hacky. And I'll take a look at that WebProvider class, it's 7 years old and full of TODO stub methods

PR #686, opened 2019-04-13
Important fixes

baka-kaba:


baka-kaba:
Fix weird problems caused by AwfulErrors, and tweak error handling

PR #696, opened 2019-09-07
Set navigationBarColor for the base themes

wardog1uk:
This fixes the navigation bar always being bright white, even in dark themes. At least on my new phone.

I think it also changes it from black to forums_blue_darker in the default theme for older phones. Wasn't sure whether to just have them both as black.

### DIFF
--- a/Awful.apk/src/main/res/values/themes.xml
+++ b/Awful.apk/src/main/res/values/themes.xml
@@ -67,6 +67,7 @@
         <item name="awfulPopUpTheme">@style/ThemeOverlay.AppCompat.Light</item>
         <item name="windowActionModeOverlay">true</item>
 
+        <item name="android:navigationBarColor">?colorPrimaryDark</item>
     </style>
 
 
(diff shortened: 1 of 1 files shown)


baka-kaba:
Ok, seems like it works ok to me - the only question is which colours should we be using. What do you think @Sereri ?

image

I guess using ``primaryDark`` is probably the best choice since it matches the status bar, but it doesn't give a "light theme" look that I think a lot of people are expecting (like a white bar with grey buttons). We always kept the bar white on black because that was the standard system look, so it never really matched the app colours outside of black themes, and it doesn't necessarily need to match them now. I'd look at the Material Design guidance but they want the nav bar to disappear anyway soooo

Either way we'll need to make that page bar match too

baka-kaba:
I'll probably tweak it so we can try out some other page bar colours - won't hurt to have it properly themed so it's easy to change in future anyway

cheers @wardog1uk !

baka-kaba:
All right I'll merge this and we can hash it out in beta, where people are hungry for change and love to give feedback

PR #713, opened 2020-06-11
AndroidX again

Sereri:


Sereri:
So I went fuck it and gave it a shot.
WORKSONMYDEVICE

Sereri:
Oh, yeah
Converts to AndroidX, replaces Crashlytics with the firebase version and for good measure I replaced the image loader that has been dead for 5 years with something else. Zoomable gifs here we goooo.

Awful.apk/build.gradle
```diff
@@ -117,22 +120,22 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
-    implementation 'com.android.support:appcompat-v7:28.0.0'
-    implementation 'com.android.support:design:28.0.0'
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
baka-kaba, line 123: There's an issue on older APIs where sp text size values are being treated as dp, so basically everything looks like the system font size is set to Normal even if it isn't.

It's fixed in appcompat-1.2.0-rc01, which we could use until the final version is out (I've been using it in something else and haven't run into any issues)

Sereri:
So in the meantime android 11 was released and 1.2 is not longer a RC.

Thanks Richard, I guess? So long, fucker.

PR #728, opened 2021-06-18
Export settings now retries when file access permissions are granted.

### LINKED ISSUE #705: Export Settings doesn't if you haven't granted file access perms
Yeah, it pops up the request and if you grant permission, you'd expect it to complete the export process. It does not! And seeing as you're probably only exporting to do a reinstall or something, finding out no backup was saved is probably going to be a problem. Right now you have to tap Export again after allowing access.

So yeah, it should repeat the export process when the user grants the permission

kahays:
Should fix #705? On my machine/test device running Android 10, I needed to change `AndroidManifest.xml` to get the fix to work reliably from a clean install, but didn't discover that until I had already pushed the other change 🤦‍♀️ sorry!

### DIFF
--- a/Awful.apk/src/main/AndroidManifest.xml
+++ b/Awful.apk/src/main/AndroidManifest.xml
@@ -31,6 +31,7 @@
         android:roundIcon="@mipmap/ic_launcher"
         android:theme="@style/Theme.AwfulTheme.Launcher"
         android:usesCleartextTraffic="true"
+        android:requestLegacyExternalStorage="true"
         >
         <uses-library android:name="org.apache.http.legacy" android:required="false" />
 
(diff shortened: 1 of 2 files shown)


Sereri:
Hey,
Sorry it took me so long. I currently really only have time for this on the weekend and the last two weekends...

Anyway, the change you sent would in theory fix the issue, yes. However as on Android version 30, the requestlegacyexternalstorage option will no longer work (see https://stackoverflow.com/questions/63364476/requestlegacyexternalstorage-is-not-working-in-android-11-api-30). We will have to use A30 as soon as November, so in around 4 months.

So while this is a solution for the very short run, this will not really help longer than that.

I've looked a bit into this and tried changing it to use getExternalFilesDir() instead but that basically dumps it into a folder most users won't be able to touch.
I guess the best would be to use the MediaStore and save it as MediaStore.Files maybe?

Can you try to refactor the code to use that?

kahays:
Absolutely, I'll try to put that together in the next couple weeks. Thanks for looking into everything :)

PR #737, opened 2022-01-16
Support alternate avatars if primary avatar dimensions are 1x1.

kahays:
People use long avatars by embedding extra images in their title and setting their avatar to a 1x1 image. This attempts to detect this and use the intended image instead. A test thread is [here](https://forums.somethingawful.com/showthread.php?threadid=3934728) (the irony is not lost on me.)

Feedback is totally welcome as I'm not well-versed in all the areas this touches.

### DIFF
--- a/Awful.apk/src/main/assets/javascript/thread.js
+++ b/Awful.apk/src/main/assets/javascript/thread.js
@@ -188,15 +188,7 @@ function processPosts(scopeElement) {
 
     // handle all GIFs that are not avatars
 	if (listener.getPreference('disableGifs') === 'true') {
-		scopeElement.querySelectorAll('img[title][src$=".gif"]:not(.avatar)').forEach(function each(gif) {
-			if (!gif.complete) {
-				gif.addEventListener('load', function freezeLoadHandler() {
-					freezeGif(this);
-				});
-			} else {
-				freezeGif(gif);
-			}
-		});
+		scopeElement.querySelectorAll('img[title][src$=".gif"]:not(.avatar)').forEach(prepareFreezeGif);
 	}
 
     // this handles all avatar processing, meaning if the avatar is a GIF we need to handle freezing as well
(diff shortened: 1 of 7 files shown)


Awful.apk/src/main/java/com/ferg/awfulapp/thread/ForumParsing.kt
```diff
@@ -121,13 +121,20 @@ class PostParseTask(
...
-            // grab the custom title, and also the avatar if there is one
+            // grab the custom title, and also avatar and alternate avatar if there are any
             postData.selectFirst(".title")!!
                 .also { put(AVATAR_TEXT, it.text()) }
-                .selectFirst("img")
-                ?.let {
-                    tryConvertToHttps(it)
-                    put(AVATAR, it.attr("src"))
+                .select("img")
+                .forEachIndexed { index, image ->
```
Sereri, line 129: Wouldn't that keep overwriting A_S if there are more than 2 images in the user title?

Awful.apk/src/main/assets/mustache/post.mustache
```diff
@@ -2,7 +2,7 @@
 	<header class="postheader">
         <div class="posterinfo">
             {{#avatarURL}}
-            <img class="avatar" src="{{.}}" title="{{username}}'s avatar" />
+            <img class="avatar" src="{{.}}" title="{{username}}'s avatar" data-avatar-second-src="{{avatarSecondURL}}"/>
```
Sereri: I'd go with `{{#avatarSecondURL}}data-avatar-second-src="{{avatarSecondURL}}" {{/avatarSecondURL}}`

Awful.apk/src/main/assets/javascript/thread.js
```diff
@@ -186,8 +186,9 @@ function processPosts(scopeElement) {
 		highlightOwnQuotes(scopeElement);
 	}
 
+    // handle all GIFs that are not avatars
 	if (listener.getPreference('disableGifs') === 'true') {
-		scopeElement.querySelectorAll('img[title][src$=".gif"]').forEach(function each(gif) {
+		scopeElement.querySelectorAll('img[title][src$=".gif"]:not(.avatar)').forEach(function each(gif) {
```
Sereri: Since the `each(gif)` and the code at l216 are mostly the same, it would make sense to put it in a function, prepareFreezeGif or whatever.
Then just check if `this.naturalWidth === 1 && this.naturalHeight === 1 && this.dataset.avatarSecondSrc` and if so change the src. After that prepareFreezeGif if disableGifs is on.

Sereri (changes requested):

Awful.apk/src/main/java/com/ferg/awfulapp/thread/ForumParsing.kt
kahays: It's the next line that makes sure it doesn't -- in a terrible hack, we only take action if the index is 0 or 1, so just the first and second images are processed.

kahays:
Thanks for reviewing :) I'll get to those changes in hopefully a couple weeks, everything in my house is packed up right now.

Awful.apk/src/main/java/com/ferg/awfulapp/thread/ForumParsing.kt
kahays: Latest commit uses take(2), clarifying the intent here.

Sereri (approved):

Sereri:
Great timing 😄, changes look good 👍 
Is there other code you want to refine or is this now good to go?

kahays:
yay! thank you 😃 I'm OK with this as it stands, no more changes planned.

PR #753, opened 2022-09-18
Fix rare pull-to-refresh bug from acting on stale data.

kahays:
As listed on the forums, there's a [rare bug where pull-to-refresh will just reload the current page](https://forums.somethingawful.com/showthread.php?threadid=3571717&pagenumber=192#post526341997) despite new pages being available. This patch basically just removes the variable holding the stale value in question.

To reproduce:
1. Load a thread where the last page is maxed out on posts.
2. Wait until there are new posts.
3. Switch away from the app, and switch back. Metadata will update and the page bar will reflect the new posts.
4. Swipe up and the page will refresh instead of loading the new page.


kahays:
Oops. This commit would only create a different bug >.<

Sereri:
Well thank dog I have glacial response times :smile:

PR #755, opened 2022-09-22
Fix GIF freezing issues.

kahays:
With the option to disable GIF animation, sometimes GIFs still don't freeze on page load ([an example](https://forums.somethingawful.com/showthread.php?threadid=3904417&pagenumber=7004&perpage=5).) Strangely, I've had no problems when the GIFs are further down in the page, like [Crusader's post here](https://forums.somethingawful.com/showthread.php?threadid=4001543&pagenumber=1551&perpage=5). Edit: my posts per page setting is 5, for reference.

With some of the changes, there may be some dead code left over, but I'm not sure if _all_ images get wrapped in `<a>` tags due to the exception in `AwfulPost.java:507` checking if an image has a `nolink` attribute. I couldn't find an example of an image with that attribute but I can't say for sure.

### DIFF
--- a/Awful.apk/src/main/assets/javascript/thread.js
+++ b/Awful.apk/src/main/assets/javascript/thread.js
@@ -187,26 +187,26 @@ function processPosts(scopeElement) {
 		highlightOwnQuotes(scopeElement);
 	}
 
-    // handle all GIFs that are not avatars
+	// handle all GIFs that are not avatars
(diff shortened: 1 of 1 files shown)


Sereri:
Last thing, the file is currently using tabs as spacers. Please convert the entire file to spaces or update your changes to use tabs, thanks 😃

kahays:
Great catch! Thank you, should be changed back to tabs now.

PR #757, opened 2022-09-26
Restore WebView scroll position after fullscreen video (uses Javascript.)

kahays:
As reported by [Flipperwaldt](https://forums.somethingawful.com/showthread.php?threadid=3571717&pagenumber=190&perpage=40#post523660176), when viewing a thread and after exiting fullscreen video, the thread page is returned to the top. Other people on StackOverflow have [had the same issue](https://stackoverflow.com/questions/61676109/webview-goes-back-to-page-top-after-exit-from-fullscreen-video).

I can't find the _right_ way to do this. The DOM `fullscreenchange` event doesn't fire when entering/exiting fullscreen, I couldn't find a callback for when the WebView is completely ready to restore a scroll position, and I couldn't get the proposed fix from StackOverflow to work reliably on my slower phone.

So this is a timeout-based approach, but using `evaluateJavascript` in the WebView, without assuming that anything from `AwfulWebView` or `thread.js` would be available. Feedback is always welcome, I'd love to find that there's a better way to fix this issue.

Note: TikToks won't go fullscreen at all, they want picture-in-picture support and that's not implemented.

Test threads:
- [OSHA in GBS](https://forums.somethingawful.com/showthread.php?threadid=3904642)
- [interesting youtubes in YOSPOS](https://forums.somethingawful.com/showthread.php?threadid=3868858)
- [PYF animated music videos](https://forums.somethingawful.com/showthread.php?threadid=3905650)
- [PYF TikToks](https://forums.somethingawful.com/showthread.php?threadid=3973214)

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/webview/LoggingWebChromeClient.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/webview/LoggingWebChromeClient.java
@@ -69,6 +69,34 @@ public class LoggingWebChromeClient extends WebChromeClient {
             callback.onCustomViewHidden();
             return;
         }
+
+        // we lose the scroll position when viewing things fullscreen.
+        //
+        // it appears to work if we store the scroll position, then
+        // immediately catch the scroll events generated when the scroll
+        // position changes, then scroll right back.
+        //
+        // for embedded videos, doesn't work when video is scrolled far
+        // enough down to hide <a class="video-link">. there's a hacky
+        // correction below.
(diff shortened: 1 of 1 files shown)


Sereri:
This seems to contain 2 issues in one (restore place after fullscreen / ignored posts) , can you split these into 2 separate PRs?

kahays:
Will do! I messed up the commit history, will fix or resubmit this as a separate issue. Thanks!

kahays:
Hopefully fixed now, let me know if it's still showing as multiple unrelated commits.

PR #761, opened 2022-10-09
Add Android 13 themed icons.

kahays:
This adds monochrome icon support for Android 13 (API 33.) Vector icons were autogenerated from black-and-white raster files, so have small rough edges but seem OK for the purpose. I can totally redo the PR if there are vector assets I should have used instead.

Warnings about Gradle/API 33 incompatibility are intentionally not suppressed, just in case we run into issues down the line and wonder what the heck. The internet suggests that problem would be really really unlikely anyway!


Sereri:
These are the files the current icons are made from:
https://drive.google.com/drive/folders/1qtVRK10kmz9gEEbSB0m56uxiOPaYXy0e

kahays:
Thanks! Looks like all the files are raster files, so I feel OK about the autogenerated vector icons. I'd be OK to accept/reject the PR as it stands :)

PR #765, opened 2022-11-12
Update shortcut creation process.

kahays:
Something in the most recent release causes [bookmark shortcuts to break](https://forums.somethingawful.com/showthread.php?threadid=3743815&pagenumber=18#post527231867), which can be replicated on API 33 by creating the bookmarks shortcut and then changing the themed icons setting. Once the setting is changed, the shortcut is permanently broken, even if the themed icons setting is changed back.

This fixes shortcut functionality, but unfortunately the shortcut icon won't reflect the themed icon setting (if applicable,) and I can't find a workaround.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/UserCPShortcutActivity.kt
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/UserCPShortcutActivity.kt
@@ -3,6 +3,9 @@ package com.ferg.awfulapp
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
+import androidx.core.content.pm.ShortcutInfoCompat
+import androidx.core.content.pm.ShortcutManagerCompat
+import androidx.core.graphics.drawable.IconCompat
 import com.ferg.awfulapp.preferences.AwfulPreferences
 
 /**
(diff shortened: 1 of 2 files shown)


Sereri:
Considering the Gmail and Youtube apps on my emulator also don't have theme working, I'm guessing it's not possible at this point. Nice oversight, Scroogle
grafik


I consider this a bugfix and merged it despite a few points I'd love to see fixed in the future:
Since this is now a full-fledged shortcut, we should get into the (pinned?) shortcut menu
Its not a widget anymore and therefore should not show up in the widget menu anymore

I think the shortcuts only work starting with API 25? Might make sense to up the min version (or cut this functionality for versions below) for this.
Its 2022, what are you doing on API 21-24 anyway.

kahays:
I'll get the shortcut working in pinned and static forms, whatever the API levels are for those.

Do we have any data on how many folks are on what API level? Seems like the widget version works all the way down to API 21, and I know old device support sucks but as a former old-phone-haver I have some sympathies.

Sereri:
Looked this up and here is everything below Android 7 (plus 7.1 because that is less than A6 users :V )
grafik
We could probably cut off everything below 8 but I don't see a benefit from that. API25 seems to be 6 years old.
I'd like to see it gone from the widgets, but I don't think it'll be workable to have that as a fallback only for API 21-24.

kahays:
Oh, that's significantly fewer than I thought, thanks for checking. I'll go ahead and remove the widget bit and see how many people complain 🤷‍♀️

PR #766, opened 2022-11-12
Set active theme based on system. (Options light/dark/battery/system.)

kahays:
Running up that hill (toward feature parity with iOS.) Tested on emulators running APIs 21, 25, 29 and 33. Tested on device running API 29. Please let me know if you encounter issues!

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
@@ -2,11 +2,15 @@ package com.ferg.awfulapp
 
 import android.annotation.SuppressLint
 import android.app.Activity
+import android.content.BroadcastReceiver
+import android.content.Context
 import android.content.Intent
+import android.content.IntentFilter
 import android.content.pm.ActivityInfo.*
 import android.graphics.Typeface
 import android.net.http.HttpResponseCache
 import android.os.Bundle
+import android.os.PowerManager
 import androidx.annotation.CallSuper
 import androidx.appcompat.app.AppCompatActivity
 import android.text.method.ScrollingMovementMethod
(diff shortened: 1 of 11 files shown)


Sereri:
Functionally this seems to work fine, aside from the thread view being reloaded when the theme switches. That might be a deal breaker in of itself.

I am also not sure about having 3 theme selection dropdowns. Especially the first one is Active theme selection could use some better description. But I'm wondering if there is generally some better way to do this. 
One solution might be to have this as a separate "Auto" theme, that just switches between the Default and Dark themes, instead of making it fully customizable. Then have battery as a toggle below that?

As for the page reload, losing your place in the thread because of time of day would suck. Have you looked into DayNight Theme?

kahays:
Thanks for reviewing! :)

I briefly looked into DayNight theming but decided against it because I was hoping to retain day/night theme configurability.

By default activities are recreated when the uiMode changes, so I thought saving the WebView scroll position before the page reload would put users close enough to where they were to not cause issues. Seems to work ok on pages without embeds/videos etc., but I do see the issues on pages with a lot of media, so *that* approach may be dead in the water.

I'll try to rework this so we handle the uiMode changes ourselves and don't need to recreate activities. I do like your idea of the "Auto" theme, in case I can't get configurability to work. For "Active theme", what would a better name be? I'm thinking of renaming the "Theme" and "Dark mode theme" options to "Light theme" and "Dark theme".

Sereri:
>  Seems to work ok on pages without embeds/videos etc., but I do see the issues on pages with a lot of media, so that approach may be dead in the water.

Ah I think the main issue for me was, that I had hidden read posts.
The other issue is of course embeds, but also that different themes might have a different place in the thread at the same page height. E.g. OLED vs Dark

As for the Active theme name, maybe it needs to to be restructured. Something like "Switch to dark theme" with "Never", "In Battery Saver Mode", "System Preference" or something like that with Never disabling/hiding the Dark theme setting.

kahays:
> The other issue is of course embeds, but also that different themes might have a different place in the thread at the same page height. E.g. OLED vs Dark

Never thought of that.. baka kaba had an idea about storing page position as offsets from individual posts, I'll check into that to see if it alleviates the biggest problems.

> As for the Active theme name, maybe it needs to to be restructured. Something like "Switch to dark theme" with "Never", "In Battery Saver Mode", "System Preference" or something like that with Never disabling/hiding the Dark theme setting.

This sounds much better, I like the idea of disabling instead of hiding for that little bit of discoverability. But if it looks cluttered hiding is good too.

PR #777, opened 2023-06-23
Add support for Cloudflare captcha challenges

### LINKED ISSUE #776: Issues with Cloudflare Captchas
In non-Western countries, the Cloudflare DDoS protection that SA is behind prompts for captchas regularly (once every 1-2 hours). In Awful, this currently means that the app can not be used without an active VPN connection to a Western country, as the captcha pages just lead to errors.

I'm not sure how exactly the captchas work, but they probably leave behind some browser state (cookies?) to identify the "blessed" session. It might be possible for Awful to detect the Cloudflare captchas, display the captcha page (probably by necessity using the same user-agent etc. as the app itself), and then persist whatever browser state change happens as a result.

tazjin:
This change implements a new activity which can display captcha challenges to users and synchronise the relevant cookie and user-agent data to make requests succeed after a captcha challenge. I made a [demo video](https://tazj.in/blobs/awful-captcha.mp4).

This fixes #776. Please see that issue for some more context.

The change is fairly complex, I suggest reviewing the individual commits in order. Each commit has a detailed description of what it does and why.

There are several issues with this implementation:

* ~~The user-agent synchronisation requires a web view to be instantiated, which can only occur inside of an activity. The only activity where this was reasonable to do is `CaptchaActivity`. This means that after a restart of the app, even if the captcha clearance is still valid, the user will briefly see a flash of the forums index page.~~

  Fixed by initialising the user-agent on app launch in the `AwfulApplication` class.
* The user always sees a flash of the forums index page after solving a captcha, which is due to the cumbersome method of detecting whether a request in a `WebView` succeeded.
* The handler for invoking the captcha challenge can not easily be installed in a central location, so it's done in some specific places (login, forums refresh, thread reload) which should cover most use-cases.
* ~~I haven't figured out how forum icons are loaded, they don't seem to be going through `AwfulRequest`, so they're not using the user-agent and cookie that they should. This means that they fail and a placeholder is displayed on captcha'd connections.~~
  
  Update: Fixed by adding `AwfulImageLoader` class.
* ~~Avatars and smilies don't load. They seem to be loaded in another kind of web view that isn't using the right settings.~~ Fixed in the web view for thread display.
* Sometimes on repeated captcha prompts, the captcha page is just white and doesn't display anything. No idea what's going on there, the Chrome inspector thing shows that it's actually a Cloudflare page that's just ... empty. This goes away after restarting the app. I have no idea what state in the app leads to this occuring.

Still, this is a lot better than before as with this change Awful goes from unusable to usable in non-Western countries (it was previously unusable in China, Russia, ...) and on connections to popular VPN providers which trigger captcha challenges. The issues listed above can be addressed in follow-up changes.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
@@ -73,8 +73,14 @@ abstract class AwfulActivity : AppCompatActivity(), AwfulPreferences.AwfulPrefer
                 if (ignoreFormkey == null || userAvatarUrl == null) RefreshUserProfileRequest(this@AwfulActivity).sendBlind()
                 if (ignoreFormkey == null) FeatureRequest(this@AwfulActivity).sendBlind()
             }
-        // TODO: this interferes with the code in AwfulFragment#reAuthenticate - i.e. it forces "return to this activity" behaviour, but fragments may want to return to the main activity.
-        // And the activity isn't always the authority, e.g. using BasicActivity which is a plain container where all the specific behaviour is handled by its fragment. It's awkward
+
+            // When a captcha is being handled, the login screen should be suppressed. Solving
+            // captchas is required for logging in, so overriding the captcha view makes the app
+            // unusuable otherwise.
+            CaptchaActivity.isCaptchaBeingHandled() -> { /* do nothing */ }
+
+            // TODO: this interferes with the code in AwfulFragment#reAuthenticate - i.e. it forces "return to this activity" behaviour, but fragments may want to return to the main activity.
+            // And the activity isn't always the authority, e.g. using BasicActivity which is a plain container where all the specific behaviour is handled by its fragment. It's awkward
             this !is AwfulLoginActivity -> showLogIn(returnToMainActivity = false)
         }
     }
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
@@ -46,18 +46,6 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Environment;
 import android.os.Handler;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-
-import com.ferg.awfulapp.search.SearchFilter;
-import com.google.android.material.floatingactionbutton.FloatingActionButton;
-import androidx.fragment.app.FragmentManager;
-import androidx.loader.app.LoaderManager;
-import androidx.core.content.ContextCompat;
-import androidx.loader.content.CursorLoader;
-import androidx.loader.content.Loader;
-import androidx.core.view.MenuItemCompat;
-import androidx.appcompat.widget.ShareActionProvider;
 import android.text.TextUtils;
 import android.text.format.Formatter;
 import android.view.InflateException;
@@ -68,7 +56,6 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.webkit.CookieManager;
-import android.webkit.CookieSyncManager;
 import android.webkit.DownloadListener;
 import android.webkit.JavascriptInterface;
 import android.webkit.WebView;
@@ -88,6 +75,7 @@ import com.ferg.awfulapp.preferences.Keys;
 import com.ferg.awfulapp.provider.AwfulProvider;
 import com.ferg.awfulapp.provider.AwfulTheme;
 import com.ferg.awfulapp.provider.ColorProvider;
+import com.ferg.awfulapp.search.SearchFilter;
 import com.ferg.awfulapp.task.AwfulRequest;
 import com.ferg.awfulapp.task.BookmarkRequest;
 import com.ferg.awfulapp.task.IgnoreRequest;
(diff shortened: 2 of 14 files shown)


tazjin:
The issue with avatars and smilies not loading beats me. If I understand correctly how this works, a thread is displayed in an `AwfulWebView`. I've added code to thread the cookies into `AwfulWebView` (not pushed as of this commit), but it doesn't seem to have any effect (though I can verify the cookies are being added).

When displaying threads, the app also logs stuff like this:

```
2023-06-23 22:37:02.453   933-1127  chromium                com.ferg.awfulapp.debug              E  [ERROR:ssl_client_socket_impl.cc(978)] handshake failed; returned -1, SSL error code 1, net_error -101
2023-06-23 22:37:02.583   933-1127  chromium                com.ferg.awfulapp.debug              E  [ERROR:ssl_client_socket_impl.cc(978)] handshake failed; returned -1, SSL error code 1, net_error -101
```

and more of these in busier threads, but I don't really know where they're coming from. Might be related? Though I don't know why an SSL issue would pop up *only* on `fi.somethingawful.com` and `i.somethingawful.com`?

The Android web view APIs are pretty sucky, it's hard to introspect what exactly is going on in there ... I'll try to debug more by adding a request interceptor and logging the shit out of it.

tazjin:
TIL you can attach Chrome DevTools to Android web views. Well, that at least let me know that setting the cookie on the web view is definitely not working at all and somehow, mysteriously, other cookies are appearing there which the web view should not know about. What the heck!?

Edit: Fixed and new commit added.

Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
```diff
@@ -73,8 +73,11 @@ abstract class AwfulActivity : AppCompatActivity(), AwfulPreferences.AwfulPrefer
                 if (ignoreFormkey == null || userAvatarUrl == null) RefreshUserProfileRequest(this@AwfulActivity).sendBlind()
                 if (ignoreFormkey == null) FeatureRequest(this@AwfulActivity).sendBlind()
             }
-        // TODO: this interferes with the code in AwfulFragment#reAuthenticate - i.e. it forces "return to this activity" behaviour, but fragments may want to return to the main activity.
-        // And the activity isn't always the authority, e.g. using BasicActivity which is a plain container where all the specific behaviour is handled by its fragment. It's awkward
+
+            CaptchaActivity.isCaptchaBeingHandled() -> { /* do nothing */ }
```
Sereri, line 80: I didn't quite understand what this is for. Can you explain?

Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
```diff
@@ -471,20 +471,25 @@ public void onDestroy() {
         getLoaderManager().destroyLoader(Constants.POST_LOADER_ID);
     }
 
-    // TODO: fix deprecated warnings
-    private synchronized void refreshSessionCookie(){
-        if(mThreadView != null){
-        	CookieSyncManager.createInstance(getActivity());
```
Sereri, line 477: If this is the only thing using CookieSyncManager, please remove the import as well :)

Sereri (commented):
Looks pretty good 👍 
Just some minor nitpicking 😉

Awful.apk/src/main/java/com/ferg/awfulapp/AwfulActivity.kt
tazjin: If the user is not logged in, without this line, the following lines will forcefully kill the captcha activity and open the login screen - making it impossible to log in (as solving the captcha is required for that).

I can add a comment explaining this, if you'd like!

Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
tazjin: Sure, I'll go through and clean up the imports!

tazjin:
Cool! I'll fix those later when I get home, currently on the go.

I've been using this daily since opening the PR and it works, with the exception of one of the issues listed in the OP which I absolutely can not figure out. It goes like this:

Solving the captcha only works *once* per application launch. If it times out, or if networks change with the application running, it needs to be force-quit and restarted. If a captcha is loaded a second time, the captcha window is completely blank and doesn't display anything. I've attached to this with the Chrome Remote Inspector thing, and it turns out the blank page *is served by Cloudflare* and is doing something in JS (that never completes, apparently).

It seems that Cloudflare is somehow detecting that it's the same app *instance*, and trying to do something smart which doesn't work for whatever reason. I haven't been able to figure out how to wipe all the state that might be involved in telling Cloudflare about this, it doesn't seem to be a cookie (wiping all cookies on a new captcha launch doesn't help).

This is 100% reproducible, and 100% fixed by force-restarting the app. It's still totally workable (as in, better than not working _at all_) but it is a bit annoying. If you have some idea what it could be, let me know!

tazjin:
Comments addressed. I recommend moving forward with this as-is, and sorting out that issue I mentioned above separately.

tazjin:
Update on the issue above, I have a new phone where that doesn't occur and it just works perfectly all of the time. It's possible that it's entirely a local issue on the other phone, either way the PR should be ready to go.

tazjin:
@Sereri Any chance we can get this merged soon? I've noticed a few posts in the Awful thread about people with unclear loading errors, I suspect they might also be getting hit by captchas, so this might affect more people than I thought initially.

I'm still using my debug builds with this change and it works fine.

PR #786, opened 2024-11-24
Add Bluesky embeds and related preference

shnrmn:


shnrmn:
I went ahead and added Bluesky embeds. I'm Normy on the forums if that is necessary. I'm not too versed with Android dev so there may be issues but it does work and I tried to match the style of the other embed functions.

Awful.apk/src/main/res/values/strings.xml
```diff
@@ -403,6 +403,7 @@
     <!-- Post Embedding page -->
     <string name="post_embedding_description">Embed elements in posts instead of just showing a link. These options may cause scroll lag or graphical glitches</string>
     <string name="inline_tweets">Tweets</string>
+    <string name="inline_bluesky">Bluesky Posts</string>
```
Sereri, line 407: Should've been Skeets smdh

Sereri (approved):
lgtm

Feel free to add yourself to `about_contributors_array` in strings.xml as well (if you want to)

shnrmn:
> lgtm
> 
> Feel free to add yourself to `about_contributors_array` in strings.xml as well (if you want to)

Done! Only thing is the emdbeds don't use a dark theme and I don't see any way to impose it. I reached out to the Bluesky people so if you'd rather wait for a response we can hold off.

Awful.apk/src/main/res/values/strings.xml
shnrmn: 😭

PR #797, opened 2026-02-22
Improve Twitter embeds

liesandfish:
The goal of this is to reduce the number of failed Tweet embeds due to rate-limiting, but I figured while I was in there I'd fancy things up for other failure cases. Doesn't change any of the visuals or user behavior of the tweets that embed fine.

First I replaced JSONP with fetch() to proper HTTP error handling and enables caching of responses across page navigations within a thread. I just realized while typing up these notes that we still issue multiple requests for the same Tweet if it appears multiple times in a page (e.g. a post and quotes) so I'm going to fix that real quick in a subsequent commit to this PR.

Tweets that fail to embed now show a styled placeholder with the error reason from Twitter's API (e.g. deleted tweets, sensitive content) instead of silently failing to embed.

Easiest way to test this is on page 1200 of the PYF Tweets thread - it has a bunch of tweets (duh), a couple deleted tweets, and a couple of sensitive content tweets, e.g.:

image

### DIFF
--- a/Awful.apk/src/main/assets/javascript/thread.js
+++ b/Awful.apk/src/main/assets/javascript/thread.js
@@ -131,9 +131,7 @@ function loadPageHtml() {
  * Initializes the newly added posts that have just been added to the container
  */
 function pageInit() {
-	document.head.querySelectorAll('.JSONP').forEach(function removeScripts(script) {
-		script.remove();
-	});
+	window.missedEmbeds = [];
 	var spoilers = document.body.querySelectorAll('.bbc-spoiler');
 	spoilers.forEach(function each(spoiler) {
 		spoiler.removeAttribute('onmouseover');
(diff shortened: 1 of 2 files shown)


Awful.apk/src/main/assets/javascript/thread.js
```diff
@@ -131,9 +131,7 @@ function loadPageHtml() {
  * Initializes the newly added posts that have just been added to the container
  */
 function pageInit() {
-	document.head.querySelectorAll('.JSONP').forEach(function removeScripts(script) {
-		script.remove();
-	});
+	window.missedEmbeds = [];
```
liesandfish, line 134: btw without this we were just growing window.missedEmbeds unbounded across pages since the webview/window obviously arent reset in those cases

liesandfish:
Hmm I'm seeing 404s that aren't for deleted videos that I need to understand better. Might just be more cases of rate limiting or... something else? Will keep looking

liesandfish:
Okay new plan, gonna kill this and use fxtwitter's API to wholly replace the Twitter embeds as an option so I can avoid the gross Twitter rate limiting that they've ramped up recently without hopefully losing any features (and maybe gaining some control over what we have now).

PR #798, opened 2026-02-22
fix being unable to report posts by Mods/IKs

### LINKED ISSUE #782: App incorrectly prevents reporting IKs/mods
Here's the code that's filtering it out of the context menu - this should only be checking Admins, not IKs/Mods.

https://github.com/Awful/Awful.apk/blob/86d5260ea252f05731ae48b0d930a7bf0e950474/Awful.apk/src/main/java/com/ferg/awfulapp/popupmenu/PostContextMenu.java#L146

liesandfish:
Fixes https://github.com/Awful/Awful.apk/issues/782 - only affects the Report button, ignore/PM/etc. remaing unchanged in their logic.

### DIFF
--- a/Awful.apk/src/main/assets/javascript/thread.js
+++ b/Awful.apk/src/main/assets/javascript/thread.js
@@ -522,7 +522,7 @@ function showPostMenu(postMenu) {
 		postMenu.getAttribute('userid'),
 		postMenu.getAttribute('lastreadurl'),
 		postMenu.hasAttribute('editable'),
-		postMenu.getAttribute('data-role') || '',
+		postMenu.getAttribute('data-role'),
 		postMenu.hasAttribute('isPlat'),
 		avatar ? avatar.getAttribute('src') : null
 	);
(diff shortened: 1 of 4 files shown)


Awful.apk/src/main/assets/javascript/thread.js
```diff
@@ -522,7 +522,7 @@ function showPostMenu(postMenu) {
 		postMenu.getAttribute('userid'),
 		postMenu.getAttribute('lastreadurl'),
 		postMenu.hasAttribute('editable'),
-		postMenu.hasAttribute('has-role'),
+		postMenu.getAttribute('data-role') || '',
```
Sereri: I think they should be null like avatar below

liesandfish:
technically the forums support you reporting yourself but the app doesn't... i didn't fix that here because i thought it might be a cleanliness thing (to keep the context menu shorter) but...

Sereri (approved):

PR #800, opened 2026-04-03
minor report changes re: astral's backend changes

liesandfish:
1. the report message form field is now called "message" not "comments" so fix that per astral
2. stop the "this post was already reported recently" toast from happening twice due to two handlers dealing with it (this was my bug from before, not really related to the backend changes)
3. when we send a request to see if the post was already reported recently, we now grab any of the warnings on the page (which right now I think is just the "this post is older than two weeks are you sure you want to report it?" one). we show that in red in the report popup (using a normal solid red color looked fine on all the themes so i didnt bother theming it)

long-term it's a bit silly now that reports use the little popup instead of normal editor (since in the browser now it's just like making a post) but i don't think anyone will ever care about that. the warnings seemed useful, though.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
@@ -796,7 +796,7 @@ public class ThreadDisplayFragment extends AwfulFragment implements NavigationEv
 	private void showReportDialog(int postId, String warning) {
 		final EditText reportReason = new EditText(this.getActivity());
 
-		String body = "Did this post break the forum rules? If so, please report it by clicking below. If you would like to add any comments explaining why you submitted this post, please do so here:";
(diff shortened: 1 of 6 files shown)


Awful.apk/src/main/java/com/ferg/awfulapp/ThreadDisplayFragment.java
```diff
@@ -789,12 +793,27 @@ public void failure(VolleyError error) {
 			}));
 	}
 
-	private void showReportDialog(int postId) {
+	private void showReportDialog(int postId, String warning) {
 		final EditText reportReason = new EditText(this.getActivity());
 
+		String body = "Did this post break the forum rules? If so, please report it by clicking below. If you would like to add any comments explaining why you submitted this post, please do so here:";
```
Sereri: Move this to the strings.xml please

Sereri (changes requested):

Sereri (approved):

PR #802, opened 2026-04-03
track the last post time for threads so we can fix the bookmarks sorting

liesandfish:
TLDR: when you're looking at a forum thread list, it's always sorted correctly (by last post date). But when you're looking at your bookmarks, it doesn't sort "live" on the PHP side - in some post a while ago astral said there's a job every minute or so that updates the bookmarks view to match the last post time. But this means if you have active threads bookmarked, when you look at your bookmarks page you'll see them sorted incorrectly (a thread with newer unread posts will appear below a different one, and then if you refresh later even with no other posts since then it'll re-order to be fixed) and that's been driving me crazy so I added this to my fork months ago. I figure it might bother other people too so I did the work to isolate it into a PR but I'd understand if this is considered too big of a change for a minor fix like this (that is a workaround for the actual forums PHP being dumb) since it has to bump the DB version.

### DIFF
--- a/Awful.apk/src/main/java/com/ferg/awfulapp/provider/DatabaseHelper.java
+++ b/Awful.apk/src/main/java/com/ferg/awfulapp/provider/DatabaseHelper.java
@@ -20,7 +20,7 @@ import com.ferg.awfulapp.thread.AwfulThread;
 public class DatabaseHelper extends SQLiteOpenHelper {
 
     private static final String DATABASE_NAME = "awful.db";
-    private static final int DATABASE_VERSION = 38;
+    private static final int DATABASE_VERSION = 39;
 
     static final String TABLE_FORUM    = "forum";
     public static final String TABLE_THREADS    = "threads";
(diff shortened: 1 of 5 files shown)


Awful.apk/src/main/java/com/ferg/awfulapp/provider/DatabaseHelper.java
```diff
@@ -212,6 +213,8 @@ public void onUpgrade(SQLiteDatabase aDb, int aOldVersion, int aNewVersion) {
             case 37:
                 dropTables(aDb, TABLE_POSTS);
                 createPostTable(aDb);
+            case 38:
+                aDb.execSQL("ALTER TABLE " + TABLE_THREADS + " ADD COLUMN " + AwfulThread.LAST_POST_DATE + " INTEGER DEFAULT 0");
```
Sereri, line 217: I mean, yeah, you can also go the slightly more complicated route :V

Sereri (approved):

